PR #15, opened 2026-02-17
Improve guidance for installing Databricks CLI when missing

pkosiec:
## Summary

This PR improves guidance for Databricks CLI installation when the skills are added via the Cursor Marketplace.
Also, it bumps the CLI version required in the skills.

### Before

image

### After

IMO the best solution - as the commands need to be run outside of the sandbox env, and sometimes also need sudo:

image

image

image

image

### DIFF
--- a/skills/databricks-apps/SKILL.md
+++ b/skills/databricks-apps/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: databricks-apps
 description: Build apps on Databricks Apps platform. Use when asked to create dashboards, data apps, analytics tools, or visualizations. Invoke BEFORE starting implementation.
-compatibility: Requires databricks CLI (>= 0.250.0)
+compatibility: Requires databricks CLI (>= v0.288.0)
 metadata:
   version: "0.1.0"
 parent: databricks
--- a/skills/databricks/SKILL.md
+++ b/skills/databricks/SKILL.md
@@ -18,14 +18,9 @@ For specific products, use dedicated skills:
 ## Prerequisites
 
 1. **CLI installed**: Run `databricks --version` to check.
-   - **If the CLI is missing or outdated (< v0.288.0): STOP. Do not proceed or work around a missing CLI. Guide the user through installation first.**
-   - Quick install:
-     - **macOS/Linux (Homebrew)**: `brew tap databricks/tap && brew install databricks`
-     - **macOS/Linux (curl fallback)**: `curl -fsSL https://raw.githubusercontent.com/databricks/setup-cli/main/install.sh | sh`
-     - **Windows (WinGet)**: `winget install Databricks.DatabricksCLI` (restart terminal after)
-   - Verify: `databricks -v`
-   - **Sandboxed environments (Cursor IDE, containers)**: Install commands write outside the workspace and may be blocked. Present the commands to the user and ask them to run in their own terminal.
-   - For edge cases (no sudo, manual install, troubleshooting): Read the [CLI Installation](databricks-cli-install.md) reference file.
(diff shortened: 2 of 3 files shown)


skills/databricks-apps/SKILL.md
```diff
@@ -1,7 +1,7 @@
 ---
 name: databricks-apps
 description: Build apps on Databricks Apps platform. Use when asked to create dashboards, data apps, analytics tools, or visualizations. Invoke BEFORE starting implementation.
-compatibility: Requires databricks CLI (>= 0.250.0)
+compatibility: Requires databricks CLI (>= v0.288.0)
```
pkosiec, line 4: I bumped the CLI version as the newest one has the new `--version` flag.

skills/databricks/SKILL.md
```diff
@@ -17,8 +17,14 @@ For specific products, use dedicated skills:
 
 ## Prerequisites
 
-1. **CLI installed**: `databricks --version`
-   - If not: see [CLI Installation](databricks-cli-install.md)
+1. **CLI installed**: Run `databricks --version` to check.
```
pkosiec, line 20: `claude-4.6-opus-high` suggested to add "a hard STOP directive plus the actual install commands inlined, so the agent never needs to follow the link for the common cases. The detailed reference file is still there for edge cases."

skills/databricks/SKILL.md
```diff
@@ -17,8 +17,14 @@ For specific products, use dedicated skills:
...
+   - **If the CLI is missing or outdated (< v0.288.0): STOP. Do not proceed or work around a missing CLI. Guide the user through installation first.**
+   - Quick install:
+     - **macOS/Linux (Homebrew)**: `brew tap databricks/tap && brew install databricks`
```
arsenyinfo: I'd rather put a link to web documentation. Having a short version of install docs adds obligations to maintain it in the longer run

skills/databricks-apps/SKILL.md
arsenyinfo: please note the field can be ignored by the agent, so don't treat it as a hard boundary

arsenyinfo (approved):

skills/databricks-apps/SKILL.md
pkosiec: Gotcha - Do you mean that I should rather revert the version change here? or does the change still make sense and we can keep it?

pkosiec:
@arsenyinfo Thanks for your review. I addressed the feedback - I thought the inline installation instructions are necessary, but after my brief tests looks like agent reads the cli installation doc properly now. So I didn't point to the URL https://docs.databricks.com/aws/en/dev-tools/cli/install - but if you think it still makes sense, I can do it 👍 

## Screenshots

image

image

image


(After the improved sandboxed section in [c094678](https://github.com/databricks/databricks-agent-skills/pull/15/commits/c0946789cabfecb442482d357a470c6a39d44da5)):
image

skills/databricks-apps/SKILL.md
arsenyinfo: worth keeping, just don't assume it will solve the version mismatch problem completely

skills/databricks/SKILL.md
pkosiec: I thought the inline installation instructions are necessary, but after my brief tests looks like agent reads the cli installation doc properly now. So I didn't point to the URL https://docs.databricks.com/aws/en/dev-tools/cli/install - but if you think it still makes sense, I can do it 👍

pkosiec:
Merging as Arseni took another look and it is LGTM from his side 👍 Thanks, Arseni!

PR #20, opened 2026-02-24
Add granular CODEOWNERS for skill directories

lennartkats-db:
## Summary
- Add per-path ownership rules so net-new skills require @lennartkats-db approval
- Apps team (`@databricks/eng-apps-devex`) owns `databricks-apps` skill, co-owns `databricks` skill and root files
- CODEOWNERS file itself owned by @lennartkats-db and @fjakobs


arsenyinfo (approved):
should we allow more contributors to `experimental` or not here yet?

fjakobs (approved):
works for me

PR #33, opened 2026-03-13
Fix Lakebase and Apps skill guidance from agent conversation analysis

pkosiec:
## Summary

Analysis of 5 agent conversations building Lakebase-backed Databricks Apps revealed consistent patterns of wasted work. This PR fixes the root causes in skill documentation.

### Issues Fixed

- **Issue 1: Agents list ALL Lakebase projects before creating** (3/5 conversations) — Added explicit "Do NOT list projects before creating" guidance to `databricks-lakebase` skill
- **Issue 2: `apps init` fails without `--version` flag** (5/5 conversations) — Documented `--version` as an optional flag on `apps init` across all three files
- **Issue 3: `apps manifest` without `--version` causes confusion** (4/5 conversations) — Documented `--version` on `apps manifest` command

### DIFF
--- a/skills/databricks-pipelines/SKILL.md
+++ b/skills/databricks-pipelines/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: databricks-pipelines
 description: Develop Lakeflow Spark Declarative Pipelines (formerly Delta Live Tables) on Databricks. Use when building batch or streaming data pipelines with Python or SQL. Invoke BEFORE starting implementation.
-compatibility: Requires databricks CLI (>= v0.294.0)
+compatibility: Requires databricks CLI (>= v0.292.0)
 metadata:
   version: "0.1.0"
 parent: databricks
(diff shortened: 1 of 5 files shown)


skills/databricks-pipelines/SKILL.md
```diff
@@ -1,7 +1,7 @@
 ---
 name: databricks-pipelines
 description: Develop Lakeflow Spark Declarative Pipelines (formerly Delta Live Tables) on Databricks. Use when building batch or streaming data pipelines with Python or SQL. Invoke BEFORE starting implementation.
-compatibility: Requires databricks CLI (>= v0.292.0)
+compatibility: Requires databricks CLI (>= v0.294.0)
```
pkosiec: @lennartkats-db @arsenyinfo How do you feel about keeping the CLI version up to date in all skills? I believe that will make it easier to maintain, however in this particular case, DB pipelines + jobs skills for sure don't require newer version.

arsenyinfo: Not sure why it is useful. It breaks a relatively rare but still legit scenario (user has older CLI and installed newer Skills via 3rd party). Is bumping for the sake of consistency really valuable?

pkosiec: Good point that it is a rare case. Alright, reverting CLI bump for irrelevant skills 👍

arsenyinfo (approved):

PR #35, opened 2026-03-22
Add manifest v2 fields: experimental, description, min_cli_version

simonfaltum:
## Why

The Databricks CLI is adding support for skill filtering by experimental status and CLI version compatibility. The manifest needs new fields to support this.

## Changes

Before: Manifest v1 with `version`, `updated_at`, and `files` per skill.

Now: Manifest v2 adds three new fields to each skill entry:
- `experimental` (bool): Whether the skill is experimental. All current skills are `false`.
- `description` (string): One-line description of the skill.
- `min_cli_version` (string): Minimum CLI version required. Empty string means no minimum.

The CLI handles both v1 and v2 manifests gracefully. Missing v2 fields use zero-value defaults (non-experimental, no description, no version requirement).

Also updated `scripts/generate_manifest.py` to emit v2 fields with a `SKILL_METADATA` dict for per-skill configuration.

## Test plan

- [x] `python3 scripts/generate_manifest.py validate` passes
- [x] Generated manifest matches expected v2 format
- [x] All existing skills present with correct file lists

### DIFF
--- a/scripts/generate_manifest.py
+++ b/scripts/generate_manifest.py
@@ -58,7 +58,7 @@ SKILL_METADATA = {
         "experimental": False,
     },
     "databricks-apps": {
-        "description": "Databricks Apps development with AppKit SDK",
+        "description": "Databricks Apps development and deployment",
         "experimental": False,
     },
     "databricks-jobs": {
--- a/scripts/generate_manifest.py
+++ b/scripts/generate_manifest.py
@@ -103,15 +103,20 @@ def generate_manifest(repo_root: Path) -> dict:
             if f.is_file()
         )
 
-        metadata = SKILL_METADATA.get(item.name, {})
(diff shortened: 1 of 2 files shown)


scripts/generate_manifest.py
```diff
@@ -52,6 +52,30 @@ def get_skill_updated_at(skill_path: Path) -> str:
     )
 
 
+SKILL_METADATA = {
+    "databricks": {
+        "description": "Core Databricks skill for CLI, auth, and data exploration",
+        "experimental": False,
+    },
+    "databricks-apps": {
+        "description": "Databricks Apps development with AppKit SDK",
```
arsenyinfo: we got some non-appkit guidance here too - should we keep the description more generic?

arsenyinfo (approved):

fjakobs (approved):

PR #39, opened 2026-03-28
Rename databricks skill to databricks-cli for clarity

dustinvannoy-db:
## Summary

Changed the name from `databricks` to `databricks-cli` to clarify use and distinguish from other skills.

No significant changes to content of skill.

## Testing
Ran the following DABs focused test manually and results were good.

### Test 1
Create a new asset bundle for my profile DEFAULT with a single Python notebook task. Set up variables for catalog and schema that differ between dev, staging,
and prod.

### Test 2
Create a DABs config for a job that runs a Jupyter notebook at notebooks/etl_pipeline.ipynb
      using serverless compute

### Test 3
Create a DABs job config for notebooks/data_processing.ipynb using a classic
      cluster with 2 workers

### Test 4
Create a job config for running src/transform.py Databricks notebook with
      a 4-worker cluster

### Test 5
Configure a DABs job to run a Python script task at scripts/load_data.py
      using serverless

### Test 6
Set up a job to run Python script src/etl_batch.py with a cluster of 3
      workers

### Test 7
Create a new Decalarative Automation Bundle (DABs) job for running a SQL query task on serverless SQL warehouse. Do this in a new file and call the job sql_job.


lennartkats-db (commented):
I'm not sure `databricks-cli` fully covers it. Maybe something like `databricks-shared` would be better, if we need to extend the name?

This was intended as a "shared" skill that should always be consumed for all work done. Kind of like our system prompt. And then it has some progressive disclosure for shared work (auth, installation, DABs, data exploration). Especially data exploration doesn't really fit the mold here since it's not a CLI activity. The CLI is a means to an end.

dustinvannoy-db:
> I'm not sure `databricks-cli` fully covers it. Maybe something like `databricks-shared` would be better, if we need to extend the name?
> This was intended as a "shared" skill that should always be consumed for all work done.

I will close this PR and try one as `databricks-core`, which seems like a good representation to me and won't conflict with other skills being distributed. Please let me know if that name should be refined.

PR #41, opened 2026-03-31
feat: add proto-first guide as databricks-apps reference

keugenek:
## Summary

Adds proto-first design guidance as reference docs inside the existing `databricks-apps` skill (per review feedback — not a standalone skill).

**Files added:**
- `skills/databricks-apps/references/appkit/proto-first.md` — 4-phase workflow: decompose → contracts → generate → implement
- `skills/databricks-apps/references/appkit/proto-contracts.md` — concrete proto↔plugin mappings for files, lakebase, jobs

**SKILL.md updated:** Added "Typed data contracts" row to the Required Reading table.

### What it covers
- Module decomposition (files, lakebase, jobs, analytics, server boundaries)
- Proto style rules (one file per boundary, ≤12 fields, snake_case)
- Proto→Lakebase DDL type mapping
- Proto→Plugin method mapping (files upload, lakebase insert, job params)
- Volume layout convention from proto messages
- Common traps (any at boundary, JSON.parse, god messages)

- [x] Examples use least-privilege permissions
- [x] Sensitive values are obfuscated
- [x] No insecure patterns introduced

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,30 +1,12 @@
 {
   "version": "2",
-  "updated_at": "2026-03-31T18:36:08Z",
+  "updated_at": "2026-03-31T11:40:34Z",
   "skills": {
-    "databricks-apps": {
-      "version": "0.1.1",
-      "description": "Databricks Apps development and deployment",
-      "experimental": false,
-      "updated_at": "2026-03-31T18:35:18Z",
-      "files": [
-        "SKILL.md",
-        "references/appkit/appkit-sdk.md",
-        "references/appkit/frontend.md",
-        "references/appkit/lakebase.md",
-        "references/appkit/overview.md",
-        "references/appkit/sql-queries.md",
-        "references/appkit/trpc.md",
-        "references/other-frameworks.md",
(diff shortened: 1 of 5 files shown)


arsenyinfo (commented):
From my understanding of current granularity, that should not be a separate skill in existing taxonomy. 
Even appkit itself is not a separate skill but a core recommendation in databricks-apps skill. This proto-first thing is an implementation best practice, i'd expect it to be part of apps skill referenced close to other core appkit practices.

keugenek:
> From my understanding of current granularity, that should not be a separate skill in existing taxonomy. Even appkit itself is not a separate skill but a core recommendation in databricks-apps skill. This proto-first thing is an implementation best practice, i'd expect it to be part of apps skill referenced close to other core appkit practices.

agree it is more like a meta-skill - how would you recommend implementing it then? (compared to current approach)

arsenyinfo (approved):

keugenek:
isaac-bot pull master and resolve any conflicts.

PR #46, opened 2026-04-02
feat(lakebase): add deploy-first workflow and schema permission docs

pkosiec:
## Summary
- Add **Deploy-First Workflow** section to the Lakebase AppKit reference explaining schema ownership and the correct order of operations (deploy → grant `databricks_superuser` → develop locally)
- Strengthen Schema Initialization wording: SP "cannot access any existing schemas or tables", not just `public`
- Add **Schema Permissions for Deployed Apps** subsection to the Lakebase CLI skill with cross-reference
- Expand troubleshooting tables in both skills with `permission denied for schema <name>` and "works locally but fails after deploy" scenarios

### DIFF
--- a/skills/databricks-apps/references/appkit/lakebase.md
+++ b/skills/databricks-apps/references/appkit/lakebase.md
@@ -134,7 +134,7 @@ export const appRouter = t.router({
 
 ## Schema Initialization
 
-**Always create a custom schema** — the Service Principal has `CONNECT_AND_CREATE` permission, which means it can create new objects but **cannot access any existing schemas or tables** (including `public`). The SP must create the schema itself to become its owner. Initialize tables on server startup:
+**Always create a custom schema** — the Service Principal cannot access any existing schemas (including `public`). It must create the schema itself to become its owner. See **`databricks-lakebase`** skill's **Schema Permissions for Deployed Apps** for the full permission model and deploy-first workflow. Initialize tables on server startup:
 
 ```typescript
 // server/server.ts — run once at startup before handling requests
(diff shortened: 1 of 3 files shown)


calvarjorge (approved):
I wonder if there is some danger here, as we're clearly leading the agent to drop a schema when some errors are found. I know that this error should typically appear when first creating the app, but if the app is already run and being used, and the agent suddenly drops the schema, could it be problematic? Is there some backups?

pkosiec:
Thanks @calvarjorge, you're right - I'll update the PR to by default not dropping the schema, and asking the user for preference 👍

skills/databricks-apps/references/appkit/lakebase.md
```diff
@@ -132,9 +132,11 @@ export const appRouter = t.router({
 });
 ```
 
+> **Deploy first!** The Service Principal must create and own the schema. Run `databricks apps deploy` before any local development. See **`databricks-lakebase`** skill's **Schema Permissions for Deployed Apps** for details.
```
arsenyinfo: I'd consider reiterating that this guidance is for App + Lakebase combo only

skills/databricks-apps/references/appkit/lakebase.md
```diff
@@ -180,6 +182,8 @@ const prisma = new PrismaClient({ adapter });
 
 ## Local Development
 
+> **CRITICAL: Deploy the app first** so the Service Principal creates and owns the database schema. If you run locally before deploying, you'll create schemas under your credentials that the SP **cannot access** after deployment. See **`databricks-lakebase`** skill's **Schema Permissions for Deployed Apps** for the full workflow and recovery steps.
```
arsenyinfo: can we enforce it in more reliable way than just asking nicely in the prompt?

arsenyinfo (approved):

pkosiec:
Thank you Arseni, fixed (improved) the mentioned areas in https://github.com/databricks/databricks-agent-skills/pull/46/changes/fe2e03c4ac35516b89c7e75683d25baefedd8bd8 👍

lennartkats-db (approved):

PR #58, opened 2026-04-28
feat(AppKit): add skill for Jobs plugin

atilafassina:
## Summary

- Adds `skills/databricks-apps/references/appkit/jobs.md` covering the AppKit Jobs plugin merged in [databricks/appkit#265](https://github.com/databricks/appkit/pull/265): scaffolding, env-based job discovery, plugin/per-job config (deferred to upstream docs), `JobHandle` API surface, the SP-by-default vs opt-in OBO gotcha (deliberately the opposite of the Files plugin), HTTP routes, resource bindings, and troubleshooting.
- Wires the new reference into `skills/databricks-apps/SKILL.md` (phase table + "When to Use What" bullet) and `skills/databricks-apps/references/appkit/overview.md` (data-pattern selector + references table).
- Regenerates `manifest.json` via `python3 scripts/skills.py`.

The reference is intentionally lean (~170 lines) and points to `npx @databricks/appkit docs Jobs plugin` for type/route signatures, per the SKILL.md rule that docs are the API authority and skill files cover anti-patterns and gotchas.

Note: top-level `skills/databricks-jobs/` (Lakeflow Jobs) is a distinct concept and is unchanged.

## Test plan

- [ ] `python3 scripts/skills.py validate` passes
- [ ] `manifest.json` lists `references/appkit/jobs.md` under `databricks-apps`
- [ ] Links from `SKILL.md` and `overview.md` resolve to the new reference
- [ ] Spot-check API claims (handle methods, env-var rules, SP/OBO defaults) against [databricks/appkit#265](https://github.com/databricks/appkit/pull/265) and [`docs/docs/plugins/jobs.md`](https://github.com/databricks/appkit/blob/main/docs/docs/plugins/jobs.md)

Made with [Orca](https://github.com/stablyai/orca) 🐋

### DIFF
--- a/skills/databricks-apps/references/appkit/jobs.md
+++ b/skills/databricks-apps/references/appkit/jobs.md
@@ -1,0 +1,141 @@
+# Jobs: Trigger Lakeflow Jobs from Apps
+
+**For full Jobs plugin API (routes, types, config options)**: run `npx @databricks/appkit docs` → Jobs plugin.
+
+Use the `jobs()` plugin when your app needs to **trigger or monitor pre-existing Databricks Lakeflow Jobs** (notebooks, Python scripts, SQL, dbt, JARs) and surface their status to users. The jobs themselves still live as regular Lakeflow Jobs in the workspace — the plugin is the typed, resource-scoped accessor that lets app code start runs, poll status, and stream completion events.
+
+The plugin is **resource-scoped**: only jobs declared via config or discovered from `DATABRICKS_JOB_*` env vars are accessible. It is not a generic Jobs SDK wrapper — to author or schedule jobs, use the `databricks-jobs` (Lakeflow) skill instead. See [`overview.md`](./overview.md) for the cross-plugin data-pattern selector.
+
+## Scaffolding
+
+```bash
+databricks apps init --name <NAME> --features jobs \
+  --set "jobs.<resourceKey>.<field>=<JOB_ID>" \
+  --run none --profile <PROFILE>
+```
+
+**Do not guess** `--set` keys — derive them from `databricks apps manifest --profile <PROFILE>` (look up the `jobs` plugin's `resources.required` entries).
+
+Multi-job and analytics+jobs are common combinations:
+
+```bash
+databricks apps init --name <NAME> --features analytics,jobs \
+  --set "analytics.sql-warehouse.id=<WAREHOUSE_ID>" \
+  --set "jobs.<resourceKey>.<field>=<JOB_ID>" \
+  --run none --profile <PROFILE>
+```
+
+Configure job IDs via environment variables in `app.yaml` (deployed) or `server/.env` (local dev):
+
+```env
+# Single-job mode → exposed under the "default" key
+DATABRICKS_JOB_ID=123456789
+
+# Multi-job mode → exposed under lowercased keys ("etl", "ml")
+DATABRICKS_JOB_ETL=123456789
+DATABRICKS_JOB_ML=987654321
+```
+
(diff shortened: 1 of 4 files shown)


skills/databricks-apps/references/appkit/jobs.md
```diff
@@ -0,0 +1,158 @@
...
+
+// Trigger and stream status until completion (async iterable, SSE-backed)
+for await (const status of etl.runAndWait({ startDate: "2025-01-01" })) {
+  console.log(status.status); // "PENDING" | "RUNNING" | "TERMINATED" | ...
+}
+```
+
+Read methods (`lastRun`, `listRuns`, `getRun`, `getRunOutput`, `getJob`) and `cancelRun` follow the same `ExecutionResult<T>` shape. Reads cache for 60s with 3 retries; `runAndWait` has a 600s ceiling.
+
+### Execution context: SP-by-default, opt-in OBO
```
pkosiec: opt-in OBO? I thought we don't support it at all?

skills/databricks-apps/references/appkit/jobs.md
pkosiec, line 1: Do you know if we could reduce the skill a bit, while still keeping a good results? Asking because I believe there are some repetitions with our appkit docs, and the agent can already access the docs 🤔 It's purely to improve maintanability.

pkosiec, line 1: BTW, I think it makes sense to have a skill to manage jobs, so that an agent can create Apps together with jobs. That's what we do for Lakebase, Model Serving plugins. WDYT?

atilafassina: I did a heavy scan on reducing recurrences. If we want to point to others, I'm happy to review.
It's currently the 2nd smallest plugin skill we have

```
  ┌───────────────────┬───────┐
  │       File        │ Lines │
  ├───────────────────┼───────┤
  │ model-serving.md  │ 104   │
  ├───────────────────┼───────┤
  │ jobs.md (was 277) │ 171   │
  ├───────────────────┼───────┤
  │ genie.md          │ 197   │
  ├───────────────────┼───────┤
  │ lakebase.md       │ 231   │
  ├───────────────────┼───────┤
  │ files.md          │ 264   │
  └───────────────────┴───────┘
```

skills/databricks-apps/references/appkit/jobs.md
atilafassina: typo. Let me trim that out

skills/databricks-apps/references/appkit/jobs.md
atilafassina: that sounds a bit more generic than regular appkit development. No?

pkosiec: yes, but it's also helpful in the AppKit-backed apps development. The idea is that you can say something like "create me a Databricks App that does x" and then agent can figure it out what to create (what resources), before creating the app - that's how it works with Model Serving endpoints or Lakebase Autoscaling projects and branches.

That's at least how I see the skills development - so feel free to confirm that with Cong 👍

atilafassina: No, I agree with you. I'm more concerned about the scope of this PR in particular.
Amending the more generic skill would broaden the surface significantly.

pkosiec: No, no, It's not a PR blocker, we can totally merge what we have here 👍 just saying that a separate PR with simple job management (CLI based, as in other skills) would be super helpful 👍

keugenek (approved):

PR #59, opened 2026-04-28
apps+lakebase: declare-resource as primary path, SQL grant as fallback, psql wrapper

jamesbroadhead:
## Summary

Stacks on top of #56. Three related fixes for what an agent should do when an AppKit Lakebase app fails on first deploy with `password authentication failed for user '<SP_UUID>'`.

The original framing of this PR (manual `databricks_create_role()` SQL as the standard step) was wrong — it was reflecting a workaround for *my* setup gap, not the platform's intended path. Pawel pointed out that the Apps platform auto-creates the SP's Postgres role at attach time **when the app declares a `database` resource** (via `--set lakebase.postgres.branch=...` and `--set lakebase.postgres.database=...` at `apps init`, which materializes as a `database:` block in the app's `resources:` in `databricks.yml`). The auth-failed error happens when that resource is missing — the platform never knows the app needs Lakebase access. The fix is to add the resource and redeploy; manual SQL is only needed when the resource form isn't an option (shared Lakebase, custom permissions, post-hoc grants).

The PR has been reframed accordingly. Changes:

**`skills/databricks-apps/SKILL.md`** — generic rule, not lakebase-specific:
- New Step 3 in the scaffolding workflow: after `databricks apps init`, verify `databricks.yml` declares every required resource from every plugin in the manifest. Same shape of error appears for missing `sql_warehouse` (403) and missing `genie_space` (CAN_RUN denied), not just `database` — so this rule belongs in the apps skill, not the lakebase one.

**`skills/databricks-apps/references/appkit/lakebase.md`** — cross-reference:
- Prerequisites callout and troubleshooting row both flipped: declare the `database` resource (re-run `apps init` with the right `--set` flags) is the primary fix; manual SQL is the fallback for shared/pre-existing Lakebases.

**`skills/databricks-lakebase/SKILL.md`** — what's still useful here:
- `databricks psql --project … -- -d databricks_postgres -f script.sql` recipe (one command vs the existing 5-line `generate-database-credential` + `PGPASSWORD` form). Useful in its own right.
- "Grant app SP for AppKit / CRUD apps" SQL block, now clearly marked as fallback to the resource-declaration path.
- Working `databricks postgres create-role` CLI form, with body-shape note (`--json '{"spec": {...}}'`, no `{"role": ...}` wrapper). The CLI body-shape gap is being addressed separately at databricks/cli#5110.
- Two troubleshooting rows: `password authentication failed for user '<UUID>'` (now points at missing-resource as the primary cause) and the `Field 'role' is required` CLI error (explains the wrapping confusion).

## Test plan
- [x] `python3.12 scripts/skills.py validate` passes
- [x] Reproduced the original failure on `e2-dogfood`: AppKit app with no `database` resource → SP fails with `password authentication failed`. Adding the resource via `databricks apps init --set lakebase.postgres.branch=... --set lakebase.postgres.database=...` and redeploying produced an auto-created SP role and the app started clean. The fallback SQL path is also tested independently.
- [ ] Reviewer confirms the Apps platform auto-grant behavior is the intended path documented elsewhere (Pawel's claim, which matches my repro).

### DIFF
--- a/skills/databricks-apps/SKILL.md
+++ b/skills/databricks-apps/SKILL.md
@@ -156,6 +156,8 @@ npx @databricks/appkit docs ./docs/plugins/analytics.md  # example: specific doc
 
 **DO NOT guess** plugin names, resource keys, or property names — always derive them from `databricks apps manifest` output. Example: if the manifest shows plugin `analytics` with a required resource `resourceKey: "sql-warehouse"` and `fields: { "id": ... }`, include `--set analytics.sql-warehouse.id=<ID>`.
 
+3. **Verify resources after init.** Open `databricks.yml` and confirm `resources.apps.<app>.resources` contains a block for **every** required resource from every plugin you included (manifest's `resources.required`). For example, `--features analytics,genie,lakebase` must produce three blocks: `sql_warehouse`, `genie_space`, **and** `database`. A missing resource means the Apps platform won't grant the SP access to that resource at deploy time, and the app will fail at runtime — typically with `password authentication failed for user '<SP_UUID>'` (Lakebase), `403` from the SQL warehouse, or `CAN_RUN denied` (Genie). Fix by re-running `init` with the missing `--set` flag, not by hand-editing the YAML — the YAML is a generated artifact and your edit will be lost the next time someone re-scaffolds.
+
 **READ [AppKit Overview](references/appkit/overview.md)** for project structure, workflow, and pre-implementation checklist.
 
 **Genie Agent Workflow** — when the user wants a Genie-powered app, do **not** start by asking for a Genie Space ID. Instead:
--- a/skills/databricks-apps/references/appkit/lakebase.md
+++ b/skills/databricks-apps/references/appkit/lakebase.md
@@ -263,10 +263,17 @@ Then create `server/.env` with the values from the endpoint response:
 PGHOST=<host from endpoint>
 PGPORT=5432
 PGDATABASE=<your database name>
-PGUSER=<your service principal client ID>
+PGUSER=<see note below>
 PGSSLMODE=require
 LAKEBASE_ENDPOINT=projects/<PROJECT_ID>/branches/<BRANCH_ID>/endpoints/<ENDPOINT_ID>
 ```
+
+> **`PGUSER` must match the credentials the AppKit dev server uses.** The Postgres role in `PGUSER` has to correspond to the principal that produced `PGPASSWORD` (the OAuth token).
+>
+> - **Default (personal Databricks profile):** AppKit's local server authenticates as your Databricks user, so `PGUSER` is your Databricks username/email. Tables created locally will be owned by your user, not the SP — that's why the deploy-first workflow exists.
+> - **Testing the deployed flow locally:** export `DATABRICKS_CLIENT_ID=<SP_CLIENT_ID>` and `DATABRICKS_CLIENT_SECRET=...` so the dev server authenticates as the SP. Then `PGUSER=<SP_CLIENT_ID>` matches.
+>
+> If `PGUSER` and the OAuth token disagree, Postgres rejects the connection with `password authentication failed for user '<UUID>'`.
 
 Load `server/.env` in your dev server (e.g. via `dotenv` or `node --env-file=server/.env`). Never commit `.env` files — add `server/.env` to `.gitignore`.
 
--- a/skills/databricks-apps/references/appkit/lakebase.md
+++ b/skills/databricks-apps/references/appkit/lakebase.md
@@ -246,7 +246,7 @@ Check the response for the `active_deployment` field. If it exists with `status.
 
 If you skip this step, the Service Principal won't own the database schema. You'll create schemas under your credentials that the SP **cannot access** after deployment. See **`databricks-lakebase`** skill's **Schema Permissions for Deployed Apps** for the full workflow and recovery steps.
 
-> **First deploy with `lakebase`:** the SP also needs a Postgres role created via `databricks_create_role()` before it can connect at all — otherwise the deployed app fails with `password authentication failed for user '<UUID>'`. Run the **Grant app SP for AppKit / CRUD apps** SQL block from the **`databricks-lakebase`** skill once after the first deploy, then restart the app.
+> **First deploy with `lakebase`:** confirm `databricks.yml` declares a `database` resource on the app (alongside `sql_warehouse`, `genie_space`, etc.). Apps platform auto-creates the SP's Postgres role only when the database is attached as an app resource — without it, the deployed app fails with `password authentication failed for user '<UUID>'`. If the resource is missing, re-run `databricks apps init` with `--set lakebase.postgres.branch=...` and `--set lakebase.postgres.database=...`; if you can't (shared Lakebase, custom permissions), use the manual SQL fallback in the **`databricks-lakebase`** skill's **Grant app SP for AppKit / CRUD apps** section.
 
 The Lakebase env vars (`PGHOST`, `PGDATABASE`, etc.) are auto-set only when deployed. For local development, get the connection details from your endpoint and set them manually:
 
@@ -278,6 +278,6 @@ Load `server/.env` in your dev server (e.g. via `dotenv` or `node --env-file=ser
 | `permission denied for schema <name>` | Schema was created by another role (e.g. you ran locally before deploying) | **Ask the user before dropping** — `DROP SCHEMA` deletes all data. See **`databricks-lakebase`** skill's **Schema Permissions for Deployed Apps** for options |
 | Works locally but `permission denied` after deploy | Local credentials created the schema; the SP can't access schemas it doesn't own | **Ask the user before dropping** — warn about data loss, then deploy first. See **`databricks-lakebase`** skill's **Schema Permissions for Deployed Apps** for options |
 | `connection refused` | Pool not connected or wrong env vars | Check `PGHOST`, `PGPORT`, `LAKEBASE_ENDPOINT` are set |
-| `password authentication failed for user '<UUID>'` | SP has no Postgres role on the branch (first deploy with `lakebase` plugin) | Run the **Grant app SP for AppKit / CRUD apps** SQL block from the **`databricks-lakebase`** skill, then restart the app |
+| `password authentication failed for user '<UUID>'` | App's `databricks.yml` is missing a `database` resource — Apps platform never auto-created the SP's Postgres role on attach | Add the missing `database` resource (re-run `databricks apps init` with `--set lakebase.postgres.branch=...` and `--set lakebase.postgres.database=...`), redeploy. Manual SQL fallback: see **`databricks-lakebase`**'s **Grant app SP for AppKit / CRUD apps** |
 | `relation "X" does not exist` | Tables not initialized | Run `CREATE TABLE IF NOT EXISTS` at startup |
 | App builds but pool fails at runtime | Env vars not set locally | Set vars in `server/.env` — see Local Development above |
--- a/skills/databricks-apps/references/appkit/lakebase.md
+++ b/skills/databricks-apps/references/appkit/lakebase.md
@@ -246,6 +246,8 @@ Check the response for the `active_deployment` field. If it exists with `status.
 
 If you skip this step, the Service Principal won't own the database schema. You'll create schemas under your credentials that the SP **cannot access** after deployment. See **`databricks-lakebase`** skill's **Schema Permissions for Deployed Apps** for the full workflow and recovery steps.
 
+> **First deploy with `lakebase`:** the SP also needs a Postgres role created via `databricks_create_role()` before it can connect at all — otherwise the deployed app fails with `password authentication failed for user '<UUID>'`. Run the **Grant app SP for AppKit / CRUD apps** SQL block from the **`databricks-lakebase`** skill once after the first deploy, then restart the app.
+
 The Lakebase env vars (`PGHOST`, `PGDATABASE`, etc.) are auto-set only when deployed. For local development, get the connection details from your endpoint and set them manually:
 
 ```bash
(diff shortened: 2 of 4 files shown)


skills/databricks-apps/references/appkit/lakebase.md
```diff
@@ -261,6 +261,8 @@ Check the response for the `active_deployment` field. If it exists with `status.
 
 If you skip this step, the Service Principal won't own the database schema. You'll create schemas under your credentials that the SP **cannot access** after deployment. See **`databricks-lakebase`** skill's **Schema Permissions for Deployed Apps** for the full workflow and recovery steps.
 
+> **First deploy with `lakebase`:** the SP also needs a Postgres role created via `databricks_create_role()` before it can connect at all — otherwise the deployed app fails with `password authentication failed for user '<UUID>'`. Run the **Grant app SP for AppKit / CRUD apps** SQL block from the **`databricks-lakebase`** skill once after the first deploy, then restart the app.
```
pkosiec: Sorry, but that doesn't seem accurate - SP gets the role automatically created once the Lakebase project is configured as an app resource. Maybe you did find some bug in the behavior that should be confirmed with the App Hosting team?

jamesbroadhead:
Pushed a correction (84e0581) — turns out the CLI is not broken, I'd been wrapping the JSON wrong. The CLI's `--json` binds to the inner `Role` object, so the working form is `'{"spec": {"identity_type": "SERVICE_PRINCIPAL", ...}}'` (no outer `{"role": ...}` wrapper).

Updated the SKILL.md to show both forms (CLI and SQL) — SQL is still preferable in practice because you can bundle role creation + grants in one psql round-trip — and reframed the troubleshooting row around the wrapping confusion rather than saying the CLI is broken.

The real CLI gap (no convenience flags for the nested `spec` fields, marked TODO in cmd/workspace/postgres/postgres.go) deserves a separate upstream PR; I'll file that on databricks/cli.

jamesbroadhead:
Big reframe pushed in 03810f6 — Pawel pointed out that the Apps platform auto-creates the SP's Postgres role when the app declares a `database` resource, so my original "manual SQL grant" framing was treating a setup gap (my `databricks.yml` was missing the resource) as the standard workflow.

Updated:
- Primary path is now declare-resource (re-run `apps init` with `--set lakebase.postgres.branch=...` + `--set lakebase.postgres.database=...`), redeploy, platform handles the grant.
- Manual SQL block is kept as the fallback for shared Lakebases / pre-existing schemas / post-hoc grants where the resource form isn't usable.
- Added a generic rule to `databricks-apps/SKILL.md` Step 3: after `apps init`, verify every required plugin resource is declared. The same failure shape (missing resource → SP can't access at runtime) hits sql_warehouse (403) and genie_space (CAN_RUN denied) too, not just lakebase, so this rule belongs in the apps skill.

Updated PR title and body to match. Diff is now 18 insertions / 14 deletions across three files.

skills/databricks-apps/references/appkit/lakebase.md
jamesbroadhead: thanks - dropped

skills/databricks-apps/references/appkit/lakebase.md
```diff
@@ -246,6 +246,8 @@ Check the response for the `active_deployment` field. If it exists with `status.
 
 If you skip this step, the Service Principal won't own the database schema. You'll create schemas under your credentials that the SP **cannot access** after deployment. See **`databricks-lakebase`** skill's **Schema Permissions for Deployed Apps** for the full workflow and recovery steps.
 
+> **First deploy with `lakebase`:** confirm `databricks.yml` declares a `database` resource on the app (alongside `sql_warehouse`, `genie_space`, etc.). Apps platform auto-creates the SP's Postgres role only when the database is attached as an app resource — without it, the deployed app fails with `password authentication failed for user '<UUID>'`. If the resource is missing, re-run `databricks apps init` with `--set lakebase.postgres.branch=...` and `--set lakebase.postgres.database=...`; if you can't (shared Lakebase, custom permissions), use the manual SQL fallback in the **`databricks-lakebase`** skill's **Grant app SP for AppKit / CRUD apps** section.
```
pkosiec, line 249: `database` is the old Lakebase Autoscaling resource, so IMO agent might be confused with such resource name.

The new (Lakebase Autoscaling) resource is `postgres`:

Example:

```
bundle:
  name: lakebs

variables:
  postgres_branch:
    description: Full Lakebase Postgres branch resource name. Obtain by running `databricks postgres list-branches projects/{project-id}`, select the desired item from the output array and use its .name value.
  postgres_database:
    description: Full Lakebase Postgres database resource name. Obtain by running `databricks postgres list-databases {branch-name}`, select the desired item from the output array and use its .name value. Requires the branch resource name.

resources:
  apps:
    app:
      name: "lakebs"
      description: "A Databricks App powered by AppKit"
      source_code_path: ./
      # Uncomment to enable on behalf of user API scopes. Available scopes: sql, dashboards.genie, files.files, serving.serving-endpoints
      # user_api_scopes:
      #   - sql

      # The resources which this app has access to.
      resources:
        - name: postgres
          postgres:
            branch: ${var.postgres_branch}
            database: ${var.postgres_database}
            permission: CAN_CONNECT_AND_CREATE

targets:
  default:
    default: true
    workspace:
      host: https://e2-dogfood.staging.cloud.databricks.com

    variables:
      postgres_branch: projects/pkosiec/branches/production
      postgres_database: projects/pkosiec/branches/production/databases/db-dmfv-24qipl4z1k

```

skills/databricks-apps/references/appkit/lakebase.md
```diff
@@ -261,11 +263,18 @@ Then create `server/.env` with the values from the endpoint response:
...
-PGUSER=<your service principal client ID>
+PGUSER=<see note below>
 PGSSLMODE=require
 LAKEBASE_ENDPOINT=projects/<PROJECT_ID>/branches/<BRANCH_ID>/endpoints/<ENDPOINT_ID>
 ```
 
+> **`PGUSER` must match the credentials the AppKit dev server uses.** The Postgres role in `PGUSER` has to correspond to the principal that produced `PGPASSWORD` (the OAuth token).
+>
+> - **Default (personal Databricks profile):** AppKit's local server authenticates as your Databricks user, so `PGUSER` is your Databricks username/email. Tables created locally will be owned by your user, not the SP — that's why the deploy-first workflow exists.
+> - **Testing the deployed flow locally:** export `DATABRICKS_CLIENT_ID=<SP_CLIENT_ID>` and `DATABRICKS_CLIENT_SECRET=...` so the dev server authenticates as the SP. Then `PGUSER=<SP_CLIENT_ID>` matches.
```
pkosiec, line 274: This isn't possible locally, we cannot get Service Principal's client secret.

skills/databricks-apps/references/appkit/lakebase.md
pkosiec, line 249: Can we avoid duplicating the apps init command and instead, point Agent to the Scaffolding section? It'll be hard to maintain so many occurrences of the init command. Thanks!

skills/databricks-lakebase/references/connectivity.md
```diff
@@ -140,6 +140,42 @@ For production apps, combine with Pattern 2's token refresh loop and SQLAlchemy
...
+conn = psycopg.connect(
+    host=endpoint_host,      # kept for TLS SNI verification
+    hostaddr=ip,             # bypasses getaddrinfo()
+    dbname="databricks_postgres",
+    user=username,
+    password=token,
+    sslmode="require",
+)
+```
+
```
pkosiec, line 178: It was a part of my previous PR but looks like it is not needed after all - could you please cherry pick your commits on top of main to ensure only your changes are added here? Thanks!

skills/databricks-apps/SKILL.md
```diff
@@ -156,6 +156,8 @@ npx @databricks/appkit docs ./docs/plugins/analytics.md  # example: specific doc
 
 **DO NOT guess** plugin names, resource keys, or property names — always derive them from `databricks apps manifest` output. Example: if the manifest shows plugin `analytics` with a required resource `resourceKey: "sql-warehouse"` and `fields: { "id": ... }`, include `--set analytics.sql-warehouse.id=<ID>`.
 
+3. **Verify resources after init.** Open `databricks.yml` and confirm `resources.apps.<app>.resources` contains a block for **every** required resource from every plugin you included (manifest's `resources.required`). For example, `--features analytics,genie,lakebase` must produce three blocks: `sql_warehouse`, `genie_space`, **and** `database`. A missing resource means the Apps platform won't grant the SP access to that resource at deploy time, and the app will fail at runtime — typically with `password authentication failed for user '<SP_UUID>'` (Lakebase), `403` from the SQL warehouse, or `CAN_RUN denied` (Genie). Fix by re-running `init` with the missing `--set` flag, not by hand-editing the YAML — the YAML is a generated artifact and your edit will be lost the next time someone re-scaffolds.
```
pkosiec, line 159: Same comment as above:
- `database` isn't the right resource name
- it would be great to point to the Scaffolding section again

skills/databricks-apps/references/appkit/lakebase.md
```diff
@@ -261,11 +263,18 @@ Then create `server/.env` with the values from the endpoint response:
...
+>
+> If `PGUSER` and the OAuth token disagree, Postgres rejects the connection with `password authentication failed for user '<UUID>'`.
```
pkosiec, line 276: Honestly, I'd rather revert changes from line 264 to 277: they don't seem to bring any benefit unless I'm mistaken?.
Maybe we can just change the line 266 and say that:
- PGUSER doesn't need to be provided when running the app locally (why? we use the currently logged user to Databricks CLI -> don't say that in the skill, no need)
- PGUSER is injected automatically when app is deployed on Databricks Apps

skills/databricks-apps/references/appkit/lakebase.md
```diff
@@ -276,5 +285,6 @@ Load `server/.env` in your dev server (e.g. via `dotenv` or `node --env-file=ser
 | `permission denied for schema <name>` | Schema was created by another role (e.g. you ran locally before deploying) | **Ask the user before dropping** — `DROP SCHEMA` deletes all data. See **`databricks-lakebase`** skill's **Schema Permissions for Deployed Apps** for options |
 | Works locally but `permission denied` after deploy | Local credentials created the schema; the SP can't access schemas it doesn't own | **Ask the user before dropping** — warn about data loss, then deploy first. See **`databricks-lakebase`** skill's **Schema Permissions for Deployed Apps** for options |
 | `connection refused` | Pool not connected or wrong env vars | Check `PGHOST`, `PGPORT`, `LAKEBASE_ENDPOINT` are set |
+| `password authentication failed for user '<UUID>'` | App's `databricks.yml` is missing a `database` resource — Apps platform never auto-created the SP's Postgres role on attach | Add the missing `database` resource (re-run `databricks apps init` with `--set lakebase.postgres.branch=...` and `--set lakebase.postgres.database=...`), redeploy. Manual SQL fallback: see **`databricks-lakebase`**'s **Grant app SP for AppKit / CRUD apps** |
```
pkosiec, line 288: Similarly as before - let's point to the Scaffolding section

skills/databricks-lakebase/SKILL.md
```diff
@@ -237,13 +236,53 @@ TOKEN=$(databricks postgres generate-database-credential $EP --profile <PROFILE>
...
+
+**Grant app SP access to synced tables** (read-only) — run as project owner after sync is ONLINE and app is deployed:
 ```sql
 GRANT USAGE ON SCHEMA public TO "<SP_CLIENT_ID>";
 GRANT SELECT ON ALL TABLES IN SCHEMA public TO "<SP_CLIENT_ID>";
 ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT ON TABLES TO "<SP_CLIENT_ID>";
 ```
-For least-privilege, consider syncing into a dedicated schema instead of `public` so the grant is scoped to synced data only.
+
+> **Default privileges caveat.** `ALTER DEFAULT PRIVILEGES` without `FOR ROLE` only applies to tables created by the role running this statement. If sync pipelines create new tables under a different role, re-run `GRANT SELECT ON ALL TABLES IN SCHEMA public TO "<SP_CLIENT_ID>"` after each new table appears, or add `FOR ROLE <pipeline_role>` once you know which role the sync runs as.
```
pkosiec, line 248: This is what the snippet above (242-246) already does, does it make sense to repeat it?

skills/databricks-lakebase/SKILL.md
```diff
@@ -237,13 +236,53 @@ TOKEN=$(databricks postgres generate-database-credential $EP --profile <PROFILE>
...
+  --profile <PROFILE>
+```
+
+> **Least privilege.** The example creates the role with default privileges only — grant database/schema/table access via the explicit `GRANT` statements above. Don't add `membership_roles: ["DATABRICKS_SUPERUSER"]` for an app SP unless broad administrative access is intentional; superuser membership lets the app role read every Lakebase database, not just its own.
+
+> **CLI body shape.** `databricks postgres create-role`'s `--json` flag binds to the inner `Role` object — fields go directly under `spec`, **not** wrapped in `{"role": ...}`. The error `Field 'role' is required and must contain at least one subfield with a non-default value` means the inner Role had no recognized fields (often because someone wrapped the body, which the CLI strips with `Warning: unknown field: role` and ships an empty body). The CLI also doesn't yet expose convenience flags like `--spec.identity-type` ([cmd/workspace/postgres/postgres.go](https://github.com/databricks/cli/blob/main/cmd/workspace/postgres/postgres.go) marks `spec` as TODO), so you must hand-craft the JSON.
```
pkosiec, line 285: I don't think we should do that:
once the Lakebase project (branch) is added as a resource to an App, the App runtime creates proper roles with proper permissions. In this PR you already corrected agent to start again from the scaffolding section if the resource is not in the `databricks.yml` (that's a very good addition).

But let's not try to workaround the whole App resource mechanism. App that uses Lakebase must define the project as an app resource - this is a strict prerequisite here.

pkosiec (commented):
I like the `databricks psql` recommendation + the `databricks.yml` resources validation 👍 Here are my few comments.

Before next round of review, please symlink the skills and try them by yourself, just to ensure everything works as expected. Thank you!

jamesbroadhead:
Closing — stale; the substantive Apps + Lakebase concerns @pkosiec raised have been overtaken by subsequent skill restructuring. Will land any salvageable pieces in a fresh, smaller PR.

PR #63, opened 2026-05-04
feat(appkit): agent-browser skill

atilafassina:
## Summary

Adds an `agent-browser-appkit` skill to instruct users/LLMs how to visualize an AppKit app in 3 scenarios

1. Local development (serving via Vite server)
2. local proxy (testing local app FE against a real production env)
3. authenticating on Databricks and storing the credentials in the `agent-browser` session

- [x] Examples use least-privilege permissions (no unnecessary `ALL PRIVILEGES`, admin tokens, or broad scopes)
- [x] Elevated permissions are explicitly called out where required
- [x] Sensitive values are obfuscated (placeholder workspace IDs, URLs, no real tokens)
- [x] No insecure patterns introduced (e.g. disabled TLS verification, hardcoded credentials)

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,17 +1,18 @@
 {
   "version": "2",
-  "updated_at": "2026-04-30T11:02:41Z",
+  "updated_at": "2026-04-30T12:02:18Z",
   "skills": {
     "databricks-apps": {
       "version": "0.1.1",
       "description": "Databricks Apps development and deployment (evaluates analytics vs synced tables data access)",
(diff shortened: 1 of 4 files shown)


keugenek (changes requested):
Does this skill suggests using Vercel's agent-browser CLI?  https://github.com/vercel-labs/agent-browser  If not, I'd consider renaming to avoid user and LLM confusion. If yes, this need further clarification and discussion

PR #72, opened 2026-05-12
serverless-migration v2: UC MLflow patterns + AutoML rewrite + failure reporting + Genie Code install

rrf-DB:
# serverless-migration v2: UC MLflow patterns + AutoML rewrite + failure reporting + Genie Code install + multi-source enumeration

## Summary

Two rounds of updates to the `databricks-serverless-migration` skill, both driven by field gaps:

### Round 1 (commit `99b3949`) — failure reporting + Genie Code install
1. Auto-file GitHub issue on unmigratable workloads. The Failure Reporting Protocol generates a pre-filled `issues/new` URL plus an optional `gh issue create` command. Decision tree simplified to offer filing for *any* unmigratable case. MUST-NOT-CONTAIN list expanded. Defensive redaction checklist runs before write, and a deterministic post-serialization regex scrub runs before the pre-filled URL is shown.
2. Document Genie Code Agent-mode install. New "Where to Run This Skill" section + `references/install-in-databricks-genie-code.md` covering all three install methods (notebook installer, shell installer, workspace-wide Databricks App), per-user vs workspace-wide paths, and the serverless-no-CLI caveat specific to this skill.

### Round 2 (commit `9ffc6c2`) — UC MLflow + AutoML rewrite + multi-source enumeration

Five new patterns added to the pattern catalog (A2, A3, M1, M2, M3) plus two new reference docs (`mlflow-uc-patterns.md`, `multi-source-enumeration.md`), driven by end-to-end migrating seven Databricks demos to serverless.

- **A2**: builtin `max(..., key=)` / `min(..., key=)` / `sorted(..., key=)` shadowed by `pyspark.sql.functions.max` (raises `TypeError: max() got an unexpected keyword argument 'key'`). Replace with `xs.sort(key=...); top = xs[0]`.
- **A3**: AutoML on serverless → inline scikit-learn `Pipeline` rewrite (`from databricks import automl` is unavailable and the dbdemos mockup fallback hits Spark Connect JSON-serialisability).
- **M1**: `mlflow.<flavor>.log_model(..., registered_model_name=...)` under UC raises `RESOURCE_DOES_NOT_EXIST`. Drop the kwarg, call `mlflow.register_model(...)` after the run and set the alias via `MlflowClient`.
- **M2**: `RegisteredModel.latest_versions` is always `None` on UC. Use `client.search_model_versions(f"name='{name}'")` + sort+index.
- **M3**: UC requires `signature=` on every `log_model`. Infer via `infer_signature(X_sample, model.predict(X_sample))` and pass to `log_model(..., signature=signature)`.

### Review fixes (commits `e8d447d`, `e475bde`, `bbd6e4d`)
- A3 example now includes `signature=` (per M3).
- `configuration-guide.md`: clarified that `environment_key` is for dependency declaration, not a compute selector. Removing cluster references is what makes a task run on serverless.
- Scrubbed internal Slack channels, personal handles, and softened the publishing flow per a public-publishing-review pass.

## Test signal

- **Regression bench** (lint + 16 industry solution accelerator workloads + 15-notebook compatibility suite + sanity) passed on 2026-06-02: 16/16 SUCCESS in 4486s, lint PASS, compat PASS, sanity PASS (9 references resolved), exit code 0.
- **End-to-end on serverless** (seven Databricks demos re-installed, migrated with this skill, run as serverless jobs): all 7 ran SUCCESS.
- **Data-equivalence comparison** (statistical + Delta time-travel + row-level): bit-identical row counts on every user-data table; all observed drift explained by demo data-generator RNG, AutoML retraining, or the required `llama-3-1-405b` → `llama-3-3-70b` LLM substitution.

## Test plan
- [x] CI lint passes on `references/*.md`
- [x] Skill loads and resolves all 9 references
- [x] Industry solution accelerator workloads migrate cleanly
- [x] 15-notebook compatibility suite passes

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,26 +4,90 @@ Skills for AI coding assistants (Claude Code, Cursor, etc.) that provide Databri
 
 ## Installation
 
-**For Claude Code:**
+Two install paths cover the **stable** skills. They install to different places
+but end up loaded by the same agents — pick whichever fits your workflow.
+
+- **Databricks CLI** writes SKILL.md files directly into each agent's skill
+  directory (`~/.claude/skills/`, `~/.cursor/extensions/<...>`, etc.).
+- **Plugin marketplaces** (Claude Code, Cursor) cache the plugin under the
+  agent's plugin directory (e.g. `~/.claude/plugins/cache/databricks-agent-skills/`);
+  the agent discovers skills from there.
+
+**Via the Databricks CLI (canonical; supports experimental skills):**
 
 ```bash
-databricks experimental aitools install
+databricks aitools install
 ```
 
-This installs skills to `~/.claude/skills/` for use with Claude Code.
+The CLI auto-detects your coding agent(s) and installs the stable skills to the
+right location:
 
-**For Cursor:**
+- **Claude Code** → `~/.claude/skills/`
+- **Cursor**, **Codex CLI**, **OpenCode**, **GitHub Copilot**, **Antigravity**
+  → their respective skill directories
 
-Run this command in chat:
+For finer control, use the `aitools skills install` subcommand directly — it
+accepts a positional skill name and an `--experimental` flag (see the
+[Experimental Skills](#experimental-skills) section).
+
+**Via the Claude Code plugin marketplace** (stable skills only — installs every
+skill under [`./skills/`](./skills/)):
+
+```text
+/plugin marketplace add databricks/databricks-agent-skills
+/plugin install databricks@databricks-agent-skills
+```
+
+**Via the Cursor plugin marketplace:**
 
 ```text
 /add-plugin databricks-skills
 ```
 
+### CLI vs plugin marketplace
+
+| | CLI | Plugin marketplace |
+|---|---|---|
+| Stable skills | ✅ (default) | ✅ |
+| Experimental skills | ✅ (with `--experimental` or by name) | ❌ |
+| Per-skill selection | ✅ (`databricks aitools install <name>`) | ❌ (all-or-nothing) |
+| Updates | `databricks aitools update` | Plugin marketplace update flow |
+| Required outside the agent | Databricks CLI v1.0.0+ | None |
+
+If in doubt, use the CLI — it's the canonical install path and the only one that
+exposes experimental skills.
+
 ## Available Skills
 
-- **databricks-apps** - Build full-stack TypeScript apps on Databricks using AppKit
-- **databricks-serverless-migration** - Migrate classic-compute workloads (notebooks, jobs, pipelines) to serverless compute. Detects compatibility blockers, provides concrete fixes, and produces an anonymized failure report you can file as a GitHub issue when migration cannot complete. Also installable inside a Databricks workspace via Genie Code Agent mode — see [the install reference](skills/databricks-serverless-migration/references/install-in-databricks-genie-code.md).
+Stable skills shipped from [`skills/`](./skills/):
+
+- **databricks-core** — CLI, authentication, profile selection, data exploration. Parent skill for all product skills.
+- **databricks-apps** — Build full-stack TypeScript apps on Databricks using AppKit.
+- **databricks-dabs** — Declarative Automation Bundles (formerly Asset Bundles) for deploying and managing Databricks resources.
+- **databricks-jobs** — Lakeflow Jobs orchestration: task types, triggers, schedules, notifications.
+- **databricks-lakebase** — Lakebase Postgres: projects, branching, autoscaling, synced tables, Data API.
+- **databricks-model-serving** — Model Serving endpoint management, AI Gateway, traffic config.
+- **databricks-pipelines** — Lakeflow Spark Declarative Pipelines (formerly DLT) for batch and streaming.
+- **databricks-serverless-migration** — Migrate classic-compute workloads to serverless compute.
+- **databricks-vector-search** — Vector Search endpoints + indexes for RAG and semantic search.
+
+## Experimental Skills
+
+The [`experimental/`](./experimental/) directory contains additional skills
+originally imported from
+[databricks-solutions/ai-dev-kit](https://github.com/databricks-solutions/ai-dev-kit)
+(now deprecated — this repo is the source of truth going forward) on a
+**best-effort basis**.
+
+- Experimental skills are **not officially supported** — they may be used, but
+  do not follow the same review / quality bar as the stable skills under
+  [`skills/`](./skills/).
+- They are **not installed by default** by `databricks aitools install`.
+  Pass `--experimental` to install all of them, or install a specific one
+  by name (with the `--experimental` flag — e.g. `databricks aitools install
+  databricks-iceberg --experimental`).
+- See [`experimental/README.md`](./experimental/README.md) for the full list
+  and caveats.
 
 ## Structure
 
@@ -39,21 +103,24 @@ skill-name/
 
 ### Adding New Skills
 
-When experimenting with new skill variations, create a "subskill" that references the main skill and adds specific guidance:
+For a narrower variation of an existing skill, create a subskill that declares
+its parent via frontmatter. This is how the stable skills are organized today
+— each product skill sets `parent: databricks-core`.
 
 ```markdown
 ---
-name: "ai-databricks-apps"
-description: "Databricks apps with AI features"
+name: "databricks-apps-chatbots"
+description: "Databricks apps with chatbot features"
+parent: databricks-apps
 ---
 
-# AI powered Databricks Apps
+# Chatbot Apps
 
-First, load the base databricks-apps skill for foundational guidance.
+**FIRST**: Use the parent `databricks-apps` skill for app development basics.
 
-Then apply these additional patterns:
-- Custom pattern 1
-- Custom pattern 2
+Then apply these patterns:
+- Pattern 1
+- Pattern 2
 ```
 
 This approach:
@@ -63,19 +130,21 @@ This approach:
 
 ### Manifest Management
 
-Sync assets and generate manifest after adding/updating skills:
+`manifest.json` is **generated** by `scripts/skills.py` from the skill directories and frontmatter. Do not edit it by hand. CI rejects manual changes via two checks: content drift (parsed dict doesn't match what `generate` would produce) and canonical form (on-disk bytes don't match `json.dumps(..., indent=2, sort_keys=True)`).
+
+Sync assets and regenerate the manifest after adding or updating skills:
 
 ```bash
 python3 scripts/skills.py
 ```
 
-Validate that assets and manifest are up to date (for CI):
+Validate that assets and manifest are up to date (used by CI):
 
 ```bash
 python3 scripts/skills.py validate
 ```
 
-The manifest is used by the CLI to discover available skills.
+The manifest is consumed by the CLI to discover available skills.
 
 ## Security
 
@@ -83,7 +152,8 @@ Please see [SECURITY](./SECURITY) for vulnerability reporting guidelines.
 
 ## Integrity
 
-All future release tags will be GPG-signed and verifiable via `git tag -v <tag>`.
+Release tags are created by the [Release workflow](./.github/workflows/release.yml)
+and map 1:1 to a published version.
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -4,90 +4,26 @@ Skills for AI coding assistants (Claude Code, Cursor, etc.) that provide Databri
 
 ## Installation
 
-Two install paths cover the **stable** skills. They install to different places
-but end up loaded by the same agents — pick whichever fits your workflow.
-
-- **Databricks CLI** writes SKILL.md files directly into each agent's skill
-  directory (`~/.claude/skills/`, `~/.cursor/extensions/<...>`, etc.).
-- **Plugin marketplaces** (Claude Code, Cursor) cache the plugin under the
-  agent's plugin directory (e.g. `~/.claude/plugins/cache/databricks-agent-skills/`);
-  the agent discovers skills from there.
-
-**Via the Databricks CLI (canonical; supports experimental skills):**
+**For Claude Code:**
 
 ```bash
-databricks aitools install
+databricks experimental aitools install
 ```
 
-The CLI auto-detects your coding agent(s) and installs the stable skills to the
-right location:
+This installs skills to `~/.claude/skills/` for use with Claude Code.
 
-- **Claude Code** → `~/.claude/skills/`
-- **Cursor**, **Codex CLI**, **OpenCode**, **GitHub Copilot**, **Antigravity**
-  → their respective skill directories
+**For Cursor:**
 
-For finer control, use the `aitools skills install` subcommand directly — it
-accepts a positional skill name and an `--experimental` flag (see the
-[Experimental Skills](#experimental-skills) section).
-
-**Via the Claude Code plugin marketplace** (stable skills only — installs every
-skill under [`./skills/`](./skills/)):
-
-```text
-/plugin marketplace add databricks/databricks-agent-skills
-/plugin install databricks@databricks-agent-skills
-```
-
-**Via the Cursor plugin marketplace:**
+Run this command in chat:
 
 ```text
 /add-plugin databricks-skills
 ```
 
-### CLI vs plugin marketplace
-
-| | CLI | Plugin marketplace |
-|---|---|---|
-| Stable skills | ✅ (default) | ✅ |
-| Experimental skills | ✅ (with `--experimental` or by name) | ❌ |
-| Per-skill selection | ✅ (`databricks aitools install <name>`) | ❌ (all-or-nothing) |
-| Updates | `databricks aitools update` | Plugin marketplace update flow |
-| Required outside the agent | Databricks CLI v1.0.0+ | None |
-
-If in doubt, use the CLI — it's the canonical install path and the only one that
-exposes experimental skills.
-
 ## Available Skills
 
-Stable skills shipped from [`skills/`](./skills/):
-
-- **databricks-core** — CLI, authentication, profile selection, data exploration. Parent skill for all product skills.
-- **databricks-apps** — Build full-stack TypeScript apps on Databricks using AppKit.
-- **databricks-dabs** — Declarative Automation Bundles (formerly Asset Bundles) for deploying and managing Databricks resources.
-- **databricks-jobs** — Lakeflow Jobs orchestration: task types, triggers, schedules, notifications.
-- **databricks-lakebase** — Lakebase Postgres: projects, branching, autoscaling, synced tables, Data API.
-- **databricks-model-serving** — Model Serving endpoint management, AI Gateway, traffic config.
-- **databricks-pipelines** — Lakeflow Spark Declarative Pipelines (formerly DLT) for batch and streaming.
-- **databricks-serverless-migration** — Migrate classic-compute workloads to serverless compute.
-- **databricks-vector-search** — Vector Search endpoints + indexes for RAG and semantic search.
-
-## Experimental Skills
-
-The [`experimental/`](./experimental/) directory contains additional skills
-originally imported from
-[databricks-solutions/ai-dev-kit](https://github.com/databricks-solutions/ai-dev-kit)
-(now deprecated — this repo is the source of truth going forward) on a
-**best-effort basis**.
-
-- Experimental skills are **not officially supported** — they may be used, but
-  do not follow the same review / quality bar as the stable skills under
-  [`skills/`](./skills/).
-- They are **not installed by default** by `databricks aitools install`.
-  Pass `--experimental` to install all of them, or install a specific one
-  by name (with the `--experimental` flag — e.g. `databricks aitools install
-  databricks-iceberg --experimental`).
-- See [`experimental/README.md`](./experimental/README.md) for the full list
-  and caveats.
+- **databricks-apps** - Build full-stack TypeScript apps on Databricks using AppKit
+- **databricks-serverless-migration** - Migrate classic-compute workloads (notebooks, jobs, pipelines) to serverless compute. Detects compatibility blockers, provides concrete fixes, and produces an anonymized failure report you can file as a GitHub issue when migration cannot complete. Also installable inside a Databricks workspace via Genie Code Agent mode — see [the install reference](skills/databricks-serverless-migration/references/install-in-databricks-genie-code.md).
 
 ## Structure
 
@@ -103,24 +39,21 @@ skill-name/
 
 ### Adding New Skills
 
-For a narrower variation of an existing skill, create a subskill that declares
-its parent via frontmatter. This is how the stable skills are organized today
-— each product skill sets `parent: databricks-core`.
+When experimenting with new skill variations, create a "subskill" that references the main skill and adds specific guidance:
 
 ```markdown
 ---
-name: "databricks-apps-chatbots"
-description: "Databricks apps with chatbot features"
-parent: databricks-apps
+name: "ai-databricks-apps"
+description: "Databricks apps with AI features"
 ---
 
-# Chatbot Apps
+# AI powered Databricks Apps
 
-**FIRST**: Use the parent `databricks-apps` skill for app development basics.
+First, load the base databricks-apps skill for foundational guidance.
 
-Then apply these patterns:
-- Pattern 1
-- Pattern 2
+Then apply these additional patterns:
+- Custom pattern 1
+- Custom pattern 2
 ```
 
 This approach:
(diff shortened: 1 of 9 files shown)


simonfaltum (dismissed):

skills/databricks-serverless-migration/references/install-in-databricks-genie-code.md
```diff
@@ -0,0 +1,87 @@
...
+
+| Resource | Link |
+|----------|------|
+| Public docs: Extend Genie Code with agent skills | https://docs.databricks.com/aws/en/genie-code/skills |
+| Public docs: Agent skills overview | https://docs.databricks.com/aws/en/agent-skills/ |
+| Notebook + shell installer | https://github.com/databricks-solutions/ai-dev-kit |
+| This skill's source | https://github.com/databricks/databricks-agent-skills/tree/main/skills/databricks-serverless-migration |
+| Agent Skills spec | https://agentskills.io/specification |
+
+For Genie Code runtime questions, the internal Slack channel is `#eng-genie-code-team`. For the workspace-wide app deployment pattern, `#india-gcc-fe-team`.
```
simonfaltum: **Blocking for a public repo.** This line ships two internal-only Slack channels (`#eng-genie-code-team`, `#india-gcc-fe-team`) and even labels one "the internal Slack channel." External users can't reach these, and it leaks internal team structure. Worth noting: this PR's own redaction policy (`SKILL.md`, "What the report MUST NOT contain") explicitly bans Slack channel references in failure reports, so the skill is violating the rule it ships. Please replace with the public support path (GitHub issues / public docs).

skills/databricks-serverless-migration/references/install-in-databricks-genie-code.md
```diff
@@ -0,0 +1,87 @@
...
+./databricks-skills/install_skills.sh --local --install-to-genie
+```
+
+Add `--profile <YOUR_PROFILE>` if your `databricks` CLI default isn't the target workspace. Requires the `databricks` CLI installed and authenticated locally.
+
+### Option 3 — Databricks App (workspace-wide)
+
+The `mcp-ai-dev-kit` app deploys once per workspace, pulls skills from a configured GitHub repo on every redeploy, and writes to the workspace-wide `/Workspace/.assistant/skills/` path so every user in the workspace sees them. Recommended if you're rolling this skill out to a team.
+
+This requires workspace admin. Ping `#india-gcc-fe-team` for the latest deploy instructions.
```
simonfaltum: `#india-gcc-fe-team` is internal, same issue as the closing line. Drop it or point to a public channel. Also confirm `mcp-ai-dev-kit` is a public component of `databricks-solutions/ai-dev-kit` (that repo is public, so likely fine) before naming it as a deploy artifact here.

skills/databricks-serverless-migration/references/failure-reporting.md
```diff
@@ -0,0 +1,162 @@
...
+  --title "[migration-skill] <final_error_category> in <failure_phase> phase" \
+  --body-file ~/.databricks-migration-skill/reports/failure-<timestamp>.json \
+  --label migration-skill
+```
+
+This works but skips the issue template's checklist sections. Prefer the browser URL when the user is unfamiliar with the contribution flow.
+
+## What we do with reports
+
+Reports are triaged by the skill maintainers (`@lennartkats-db @simonfaltum` plus the migration-skill code owners) and used to:
```
simonfaltum: Minor leak: naming individual employees (`@lennartkats-db @simonfaltum`) in a public skill doc. This PR's own report redaction list (`SKILL.md`) bans "Databricks employee names." Prefer CODEOWNERS or a generic "migration-skill code owners" rather than personal handles.

skills/databricks-serverless-migration/SKILL.md
```diff
@@ -594,34 +705,85 @@ Write a JSON file to `~/.databricks-migration-skill/reports/failure-<ISO-timesta
...
-- **No credentials** — no secret scope names, API keys, or connection strings
-- **No data descriptions** — no column value samples, row counts tied to specific tables, or data shape details beyond the `notebook_characteristics` fields
+- **No identifiers** — no table names, column names, catalog names, schema names, secret scope names, user emails, workspace IDs, or account IDs
+- **No internal Databricks references** — no Databricks employee names, internal codenames (e.g., product code names not in public docs), `go/` links, Confluence page IDs, Google Doc IDs, Slack user or channel IDs (`U…`, `C…`), PROD-* / SEV-* / SC-* ticket numbers
+- **No customer references** — no company names, product names of customer systems, or anything that would identify the workspace's owning organization
+- **No credentials** — no tokens, API keys, connection strings, JDBC URLs, or service principal IDs
+- **No data descriptions** — no column value samples, row counts tied to specific tables, or schema fingerprints beyond the `notebook_characteristics` fields
+
+### Anonymization safety pass
```
simonfaltum, line 720: The redaction design is careful (pattern IDs only, hashed stack frames, constrained `notebook_characteristics`), but it's all performed by the model at generation time. This checklist is a set of regexes the agent is asked to apply by hand. Recommend running these regexes as a deterministic scrub over the serialized JSON before it is shown or filed, rather than trusting the model to catch every catalog/schema/email/workspace-URL. The outbound path is a public GitHub issue, so a single miss is a public leak of customer identifiers.

skills/databricks-serverless-migration/SKILL.md
```diff
@@ -594,34 +705,85 @@ Write a JSON file to `~/.databricks-migration-skill/reports/failure-<ISO-timesta
...
+| `\bgo/[a-z0-9-]+\b` | go/ links |
+| `\b(PROD|SEV|SC|JIRA)-\d+\b` | Internal ticket IDs |
+| `[0-9a-f]{20,}` (heuristic) | Likely doc/file/workspace IDs |
+| Catalog/schema/table name literals from the analyzed notebook | Drop and replace with `"<redacted>"` |
+
+The `notebook_characteristics` fields are the only safe surface for workload metadata. Do not add new fields without expanding this checklist.
+
+### After generating the report — REQUIRED output template
+
+**This is not optional.** When the decision tree above says "offer to file", you MUST produce all three of these in your final response to the user, in order:
```
simonfaltum: This pushes hard toward publishing ("not optional", "you MUST produce" the pre-filled public issue URL). Combined with model-performed redaction (see the anonymization pass above), the default nudges the agent toward emitting a one-click link that posts customer-derived data to a public tracker. Suggest making the local report the default output, with the pre-filled URL and `gh` command appearing only after the user explicitly confirms they reviewed redaction. Keep "never auto-submit", just don't lead with the publish step.

skills/databricks-serverless-migration/references/multi-source-enumeration.md
```diff
@@ -0,0 +1,68 @@
...
+
+Be permissive in the parse: parse `bundle_config.py` as Python (read its top-level assignments), then scan all top-level dict / list literals for any value that looks like a git URL (`https://github.com/...`, `git@github.com:...`) or a cloud storage path (`s3://`, `abfss://`, `gs://`).
+
+## Procedure
+
+1. **Parse `bundle_config.py`** for any upstream-source declarations.
+2. **For each upstream source**:
+   - **Git URL**: `git clone <url>` into a sibling location (e.g., `~/.databricks-migration-skill/scratch/<run-id>/external/<repo-name>/`). Honor the declared branch/ref.
+   - **S3/ADLS path**: download via `databricks fs cp -r <path> <local>` into the same sibling location.
+   - **Workspace path**: export via `databricks workspace export-dir <path> <local>`.
```
simonfaltum: Two high-risk behaviors to gate here:

1. "parse `bundle_config.py` as Python (read its top-level assignments)" should mandate static AST parsing (`ast.parse` / `ast.literal_eval`) and explicitly forbid `exec`/`import` of the file. It is workload-supplied (possibly untrusted) code, and importing it to read assignments is arbitrary code execution on the host.
2. The procedure then auto-runs `git clone`, `databricks fs cp -r`, and `workspace export-dir` on any URL or path it finds, with "be permissive in the parse." That is auto-fetching attacker-influenceable locations with no confirmation (SSRF / unintended or oversized fetches). Recommend an explicit user-confirmation gate before any clone or fetch, and restricting to declared fields (`git_source`, etc.) rather than "any value that looks like a URL."

skills/databricks-serverless-migration/SKILL.md
```diff
@@ -159,6 +237,8 @@ Scan the code for patterns that are incompatible with the serverless compute arc
...
 | LZO format files | Blocker | Convert to Parquet or Delta |
+| AutoML-trained model loaded via `mlflow.pyfunc.spark_udf(..., env_manager='local')` inside an SDP `.py` library file (E1) | Blocker | The SDP serverless image does not ship `databricks-automl-runtime`; cloudpickle.load raises `ModuleNotFoundError: No module named 'databricks.automl_runtime'`. Auto-emit `%pip install -q databricks-automl-runtime` as the first non-comment line of the SDP `.py` library file. `%pip install` is supported in SDP `.py` library files and runs once per update before SQL flows are planned. Same fix works for non-SDP notebooks loading AutoML-trained models. |
+| AutoML → sklearn rewrite (A3) with pre-existing model-serving endpoint (E2) | Blocker on first redeploy | The rewrite changes model signature (e.g., drops `id` from inputs). A pre-existing endpoint pinned to the old AutoML signature fails create/update with HTTP 400 `Failed to enforce schema of data ... Model is missing inputs ['id']`. Flip the downstream serving notebook's `force_update = False` → `force_update = True` so the endpoint re-binds to the current `prod` (or `Champion`) alias. |
```
simonfaltum: Minor: flipping `force_update = False -> True` forces a model-serving endpoint to redeploy and re-bind. Fine against a migrated copy, but if the agent applies this to a live endpoint it is a mutating, potentially disruptive action. Worth scoping the instruction to the migrated/non-prod target and confirming before forcing an endpoint update.

simonfaltum (changes requested):
Reviewed for internal leaks, data-exfiltration surfaces, and high-risk agent behavior. Flagging up front that this is a **public** repo, so everything in these files (and the PR description) is public.

**Blocking**
- Internal-only Slack channels shipped in `install-in-databricks-genie-code.md` (`#eng-genie-code-team`, `#india-gcc-fe-team`). See inline.

**Should fix before merge**
- Report redaction is performed by the model at generation time. Recommend a deterministic regex scrub over the JSON before it can be filed, since the outbound path is a public GitHub issue.
- Soften the "you MUST emit the pre-filled public issue URL" framing so local-only is the default until the user confirms redaction.
- `multi-source-enumeration.md` tells the agent to auto-clone/fetch arbitrary URLs parsed from a `.py` config. Mandate AST-only parsing (no `exec`/`import`) and a confirmation gate before any fetch.

**Minor**
- Personal maintainer handles in `failure-reporting.md` (prefer CODEOWNERS). Note this PR's own MUST-NOT-CONTAIN list bans employee names and Slack channels, which the new docs reintroduce.
- `force_update = False -> True` (E2) can disrupt a live serving endpoint. Scope it to the migrated target.

**Not inline (not in the diff):** the PR description leaks internal names/infra ("Prashanth", "SLanalytics test-stand", "ISA workloads"). PR descriptions on a public repo are public too, please scrub.

Net: nothing exfiltrates by default and the failure-report schema is well constrained. The residual data risk is entirely in model-trusted redaction plus the push to publish.

rrf-DB:
Thanks for the thorough pass. Addressed in `bbd6e4d`:

**Blockers**
- Dropped both Slack channels from `install-in-databricks-genie-code.md` and replaced with "file a GitHub issue on the public skill repo" for runtime questions, and "ask the workspace admin who owns Databricks Apps for your workspace" for the app-deploy path.
- Replaced `@lennartkats-db @simonfaltum` in `failure-reporting.md` with a CODEOWNERS reference.
- Scrubbed the PR description (the "Prashanth", "SLanalytics", "ISA" references are now "15-notebook compatibility suite", "regression bench", "industry solution accelerator workloads").

**Should-fix**
- Added a deterministic post-serialization regex scrub: after the JSON is written, the same MUST-NOT-CONTAIN pattern set is re-applied as a literal text search; if any pattern matches, the pre-filled URL is **not** shown and the user is told to redact the file manually.
- Softened the publishing CTA. Default is now local-only with a "open the file, confirm it looks right, then tell me to share it" prompt. The pre-filled URL and `gh` command appear only after explicit user confirmation, in a separate turn.
- `multi-source-enumeration.md`: parsing is now AST-only (`ast.parse` + `ast.literal_eval` on literal RHS), explicitly forbidding `exec` / `eval` / `import` of the workload file. Added an explicit user-confirmation gate before any `git clone` / `databricks fs cp` / `databricks workspace export-dir`.

**Minor**
- `E2 force_update` scoped to the migrated test endpoint: explicit "confirm the endpoint name matches the migrated copy, not a production endpoint serving real traffic" precondition before flipping.

Also ran `/external-content-review` over the whole skill bundle (skill files + 9 references). No Photon multipliers / Selective Photon / AI Runtime / internal codenames / go links / employee emails / Grafana links found. Three RED items it surfaced were the same three you flagged.

README.md
```diff
@@ -4,90 +4,26 @@ Skills for AI coding assistants (Claude Code, Cursor, etc.) that provide Databri
...
-- **Plugin marketplaces** (Claude Code, Cursor) cache the plugin under the
-  agent's plugin directory (e.g. `~/.claude/plugins/cache/databricks-agent-skills/`);
-  the agent discovers skills from there.
-
-**Via the Databricks CLI (canonical; supports experimental skills):**
+**For Claude Code:**
 
 ```bash
-databricks aitools install
+databricks experimental aitools install
```
simonfaltum: **Blocker — looks like an unintended README revert.** This branch's README predates a couple of recently merged changes, so merging it would roll them back:

- Install command goes back to `databricks experimental aitools install`. The `experimental` form was dropped as deprecated (#101); current `main` and the other skills use `databricks aitools install`.
- Re-adds the "All future release tags will be GPG-signed…" line that #116 just removed.
- Drops 7 skills from Available Skills (core, dabs, jobs, lakebase, model-serving, pipelines, vector-search) and deletes the CLI-vs-marketplace section.
- Softens the "`manifest.json` is generated — do not hand-edit" warning.

I checked the merge base (03d0cb2): it already has the full README with `databricks aitools install` and no GPG line, so this is almost certainly a rebase artifact (the PR description doesn't mention README changes).

Suggested fix — restore main's version and re-add just the one line you want:

```
git checkout origin/main -- README.md
```

then add back the serverless-migration entry that points at the Genie Code install reference.

skills/databricks-serverless-migration/SKILL.md
```diff
@@ -21,6 +21,15 @@ Analyze existing Databricks code for serverless compute compatibility and guide
 - Troubleshooting serverless-specific errors after migration
 - Choosing between Performance-Optimized and Standard mode
 
+## Where to Run This Skill
+
+This skill is published as an Agent Skill (agentskills.io) and runs in any compatible client:
+
+- **Claude Code, Cursor, or any agentskills.io client on your laptop** — the default. Install via `databricks experimental aitools install` or follow the per-client docs.
```
simonfaltum: The `experimental` form is deprecated here. Current `main` and the sibling skills use `databricks aitools install` (see `databricks-jobs/SKILL.md` and `databricks-pipelines/SKILL.md`), and the Databricks docs call it the Skills CLI. Suggest dropping `experimental`:

> Install via `databricks aitools install`

skills/databricks-serverless-migration/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: databricks-serverless-migration
-description: "Migrate Databricks workloads from classic compute to serverless compute. Use when migrating from classic to serverless, checking serverless code compatibility, or writing new serverless-compatible notebooks and jobs. Scans code for compatibility issues, provides concrete fixes for the serverless Spark Connect architecture, and guides the full migration. Not for classic DBR version upgrades or cluster configuration changes within classic compute."
+description: "Migrate Databricks workloads from classic compute to serverless compute. Scans code for serverless compatibility issues, provides concrete fixes for the serverless Spark Connect architecture, and guides the full migration to serverless environments. Use for classic-to-serverless migrations, serverless code compatibility checks, or writing new serverless-compatible notebooks and jobs. Not for classic DBR version upgrades or cluster configuration changes within classic compute."
```
simonfaltum: Heads up — this rewording quietly breaks the generated marketplace blurb. `scripts/skills.py` derives the short description by cutting at the first `. Use when` / `. Use this` / `. Triggers` / `. ALWAYS`; if it finds none, it hard-truncates at 197 chars. Moving the trigger into the middle and changing `Use when` → `Use for` means it now falls through to a mid-word cut — the manifest currently ends `…architecture, and g...`. The old wording cut cleanly to `Migrate Databricks workloads from classic compute to serverless compute.`

`scripts/skills.py validate` passes (the manifest matches the generator), so CI won't catch this. Suggest keeping the `<one-sentence summary>. Use when …` shape, with `Use when …` right after the first sentence, so the blurb cuts cleanly.

skills/databricks-serverless-migration/references/failure-reporting.md
```diff
@@ -0,0 +1,162 @@
...
+  --title "[migration-skill] <final_error_category> in <failure_phase> phase" \
+  --body-file ~/.databricks-migration-skill/reports/failure-<timestamp>.json \
+  --label migration-skill
+```
+
+This works but skips the issue template's checklist sections. Prefer the browser URL when the user is unfamiliar with the contribution flow.
+
+## What we do with reports
+
+Reports are triaged by the skill maintainers listed in the repo's [CODEOWNERS](https://github.com/databricks/databricks-agent-skills/blob/main/CODEOWNERS) file, and used to:
```
simonfaltum: This link 404s — `CODEOWNERS` lives at `.github/CODEOWNERS`, not the repo root. Should be:

`https://github.com/databricks/databricks-agent-skills/blob/main/.github/CODEOWNERS`

skills/databricks-serverless-migration/references/install-in-databricks-genie-code.md
```diff
@@ -0,0 +1,87 @@
...
+- **Genie Code's underlying model is not publicly specified** and may not be Claude. The skill is authored against Claude Code defaults; behavior parity is not guaranteed. Smoke-test on a non-critical workload first.
+- **Skills are knowledge + scripts, not arbitrary tools.** Genie Code skills can't make network calls outside the workspace boundary. The skill's GitHub-issue-filing flow ([Failure Reporting](failure-reporting.md)) still works because it generates a URL the user clicks rather than auto-submitting.
+- **File-edit sandboxing**: Genie Code can write files in the workspace, so the migration outputs (migrated notebooks, updated job JSON, env spec YAML) land where you expect.
+
+## Pointers and follow-ups
+
+| Resource | Link |
+|----------|------|
+| Public docs: Extend Genie Code with agent skills | https://docs.databricks.com/aws/en/genie-code/skills |
+| Public docs: Agent skills overview | https://docs.databricks.com/aws/en/agent-skills/ |
```
simonfaltum: Both pages exist (verified), but these are the only `docs.databricks.com/aws/en/…` URLs in the skill — everything else uses the cloud-neutral `docs.databricks.com/en/…`. Since the skill is cloud-agnostic, suggest dropping the `/aws/` prefix so it doesn't pin AWS.

skills/databricks-serverless-migration/SKILL.md
```diff
@@ -68,6 +77,31 @@ Collect the full picture of what needs to migrate to serverless:
...
+
+If the workload spans more than one user notebook (exclude `_resources/` setup notebooks from the count), process them **one at a time** rather than all at once. The agent's context window is finite, and trying to hold the full source of an 8-notebook job in active context while doing analysis, fixes, and migration triggers autocompact thrashing (Claude Code) or equivalent context-overflow failures in other clients.
+
+**Procedure**:
+
+1. **Enumerate first.** List every user notebook with its path. Do not read the bodies in full yet beyond a few lines for orientation.
+   - **Check for `bundle_config.py` (P1, H4).** If the source tree contains a `bundle_config.py`, parse it and follow any upstream-source declarations (git URLs, S3 paths, `init_scripts`, `git_source`, or custom upstream-repo declarations). Clone or fetch each referenced repo into a sibling location and include its notebooks in the enumeration. List "external sources" as a first-class artifact category in the migration plan. Without this step, multi-source demos (e.g., `dbt-on-databricks` references the upstream `dbt-databricks-c360` repo) get mis-enumerated and the real task notebooks are missed.
```
simonfaltum: Minor / question: could you clarify what the `P` codes refer to? `H4` maps to the Category H table, but I couldn't find a `P1` defined anywhere, and `P2` shows up as a pattern in the troubleshooting table too. What is `P2` (and is `P1` the same kind of thing)? A one-line definition — or just dropping the prefix — would make these resolvable for an agent following along.

skills/databricks-serverless-migration/references/multi-source-enumeration.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Procedure
+
+1. **Parse `bundle_config.py`** via `ast.parse` for any upstream-source declarations (see safety note above).
+2. **Show the parsed external sources to the user and require explicit confirmation before any fetch.** Print every git URL, S3/ADLS path, and workspace path you found, and wait for the user to approve. Do not auto-fetch without approval — these URLs come from the workload's own config and may point anywhere.
+3. **For each user-approved upstream source**:
+   - **Git URL**: `git clone <url>` into a sibling location (e.g., `~/.databricks-migration-skill/scratch/<run-id>/external/<repo-name>/`). Honor the declared branch/ref.
+   - **S3/ADLS path**: download via `databricks fs cp -r <path> <local>` into the same sibling location.
+   - **Workspace path**: export via `databricks workspace export-dir <path> <local>`.
+3. **List external sources as a first-class artifact category** in the migration plan, alongside notebooks, jobs, pipelines. Use a category like `external_sources[]` with each entry containing `source_type` (git / s3 / workspace), `url_or_path`, `ref` (if git), and `local_path` (where you put it).
```
simonfaltum: Tiny numbering nit: this is the second `3.` in the list (the step just above is also `3.`), so the procedure reads 1, 2, 3, 3, 4, 5. Looks like this should be `4.`, with the following two bumped to 5 and 6.

skills/databricks-serverless-migration/SKILL.md
```diff
@@ -594,34 +705,110 @@ Write a JSON file to `~/.databricks-migration-skill/reports/failure-<ISO-timesta
...
+
+**Never transmit the report automatically.** The user owns their data and must review before sharing. If the user declines, do not press them — log the local report path and move on.
+
+## Success report
+
+If the migration was successful, generate a report as well, using similar methods. For each workload migrated, show what was done.
+
+## Multiple workload migration
+
+When migrating multiple workloads in one session, parallelize this work as much as possible to migrate multiple workloads in parallel, including using subagents.
```
simonfaltum: Maybe it's better if we phrase these two a little more concretely — a couple of gentle suggestions:

- **Success report**: "using similar methods" is a bit vague next to the detailed failure protocol. Maybe something like: "When migration succeeds, generate a per-workload report with the same structure as the failure report (minus the redaction-to-publish step), listing the patterns detected and the fixes applied for each workload."
- **Multiple workload migration**: "parallelize this work… in parallel" reads a little redundant, and it sits right after the one-notebook-at-a-time guidance, so it's worth being explicit about scope. Maybe: "When several independent workloads are in scope, migrate them in parallel (e.g. one subagent per workload). Keep the per-notebook steps within a single workload sequential, as in Step 1.5."

Just suggestions — phrase however reads best to you.

simonfaltum (commented):
Reviewed the v2 changes — solid, clearly field-driven work. A few things I verified and liked: the failure-reporting safety (local-by-default, explicit confirm before the pre-filled URL, deterministic post-serialization scrub), the AST-only `bundle_config.py` parsing with confirm-before-fetch, and the `force_update` flip scoped to the migrated test endpoint. I also checked the Genie Code install premise against the live docs — both pages exist and the `.assistant/skills/` paths match. `python3 scripts/skills.py validate` passes.

One blocker before merge: the README changes look like an unintended rebase revert (deprecated install command, re-added GPG claim, dropped skills) — details inline. Beyond that, a few small fixes (deprecated install command in SKILL.md, the marketplace blurb now truncating mid-word, a broken CODEOWNERS link) and some nits.

Findings are inline; each is prefixed with rough severity.

rrf-DB:
Thanks Simon, all addressed in `ef558a7`. Specifically:

1. **README revert** — restored from `origin/main` (you nailed it: was a rebase artifact). The serverless-migration entry was already in main's version, so nothing to re-add.
2. **`experimental aitools install`** — dropped to `databricks aitools install` in SKILL.md:28.
3. **Description / marketplace blurb** — restored the "<one-sentence summary>. Use when …" shape so the blurb cuts cleanly. The auto-generated manifest description is now `Migrate Databricks workloads from classic compute to serverless compute.` (72 chars), back to the clean form. Verified by re-running `python3 scripts/skills.py`.
4. **CODEOWNERS link** — fixed to `.github/CODEOWNERS`.
5. **`/aws/en/` URLs** — both dropped to `/en/` so the skill stays cloud-agnostic.
6. **`P1` / `P2` codes** — both renamed:
   - `P1` → `H4` (Category H "Job-level config / multi-source" already covers `bundle_config.py`).
   - `P2` → `M4`, with a new Category M row in the pattern catalog defining it (binary-classifier `float64` → `Integer` cast for the AutoML → sklearn rewrite).
7. **Multi-source numbering** — fixed (was 1, 2, 3, 3, 4, 5 → now 1, 2, 3, 4, 5, 6).
8. **Success report + multi-workload sections** — replaced with your suggested phrasing verbatim. Both now reference the existing failure-report structure / Step 1.5 ordering, so the contracts are concrete.

Lint + sanity re-run locally, both PASS. Marketplace blurb re-verified.

Thank you for the thorough pass.

simonfaltum (approved):

PR #68, opened 2026-05-07
Improve skills based on an external benchmark for DevHub prompts

pkosiec:
## Summary

Skill improvements based on the internal report — a benchmark of Claude Opus 4.6 vs Codex 5.2 across 25 DevHub templates. The report identified error recovery patterns, troubleshooting gaps, and factual errors in skills.

### Report-driven additions
- **Token passthrough error** — added to Common Errors with correct fix (workspace admin enables user authorization)
- **Lakehouse Sync** — documented as UI-only, added REPLICA IDENTITY prerequisite, fixed Azure support status, added Postgres 17 requirement and limitations
- **Off-platform TypeScript** — added `@databricks/lakebase` standalone pattern in connectivity.md
- **CLI fallback** — added REST API exception for sandboxed environments
- **PostgreSQL extensions** — brief note with link to official docs

### Bug fixes (verified against AppKit source)
- **Lakebase API pattern** — replaced outdated `createLakebasePool()` + `pool.query()` with `AppKit.lakebase.query()` plugin pattern
- **ORM integration** — updated to use `AppKit.lakebase.pool` / `getOrmConfig()`
- **Genie resource** — restored correct `genie_space_name` variable (confirmed via `apps init` output)
- **Model serving** — replaced verbose streaming/AI Gateway sections with AppKit docs pointer; noted AI Gateway (beta) endpoints not directly supported

### False positives from initial analysis (reverted)
- Scaffolding "Go template bug" — intentional AppKit template design, not a bug
- OBO vs SP-only guidance — misleading since `apps init` handles scopes correctly
- Verify Deployment / Deployment Recovery sections — redundant (`apps deploy` already reports status)
- CLI version bump to v0.296.0 — unjustified
- Jobs/Pipelines troubleshooting tables — not report-driven
- Unity Catalog skill — separate effort
- `databricks apps logs` PAT error — actually about querying app endpoints, not the logs command

## Test plan
- [x] `python3 scripts/skills.py validate` passes
- [x] No real credentials or workspace IDs in examples
- [x] Key claims verified against official Databricks docs and AppKit source

Co-authored-by: Isaac

---

## JIRA

Related:
- [LKB-12465](https://databricks.atlassian.net/browse/LKB-12465) — Cookbook codegen: agent generates wrong AppKit API signatures
- [LKB-12428](https://databricks.atlassian.net/browse/LKB-12428) — Mode B: `as unknown as` double-assertions (Done)
[...]

PR #75, opened 2026-05-12
Remove updated_at fields from manifest

pkosiec:
## Summary

- Removes `updated_at` fields from both the top-level manifest and per-skill entries
- Deletes the `get_skill_updated_at()` function and `datetime` import from `scripts/skills.py`
- Cleans up the `normalize_manifest()` function which was stripping these fields during validation anyway

## Motivation

[Review comment on PR #68](https://github.com/databricks/databricks-agent-skills/pull/68#pullrequestreview-4253850254) from @keugenek pointed out these fields are unnecessary. The validation logic already excluded `updated_at` during comparison, confirming they served no functional purpose — just diff churn on every `scripts/skills.py generate` run. Git history is the real source of truth for when things changed.

## Test plan

- [x] `python3 scripts/skills.py generate` produces manifest without `updated_at`
- [x] `grep updated_at manifest.json` returns no matches

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,12 +1,10 @@
 {
   "version": "2",
-  "updated_at": "2026-05-22T20:18:49Z",
   "skills": {
     "databricks-apps": {
       "version": "0.1.2",
       "description": "Databricks Apps development and deployment (evaluates analytics vs synced tables data access)",
       "repo_dir": "skills",
-      "updated_at": "2026-05-22T15:54:04Z",
       "files": [
         "SKILL.md",
         "agents/openai.yaml",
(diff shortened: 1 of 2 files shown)


simonfaltum:
Isn't it nice though to have a way to easily get an overview of when a skill was last updated, both for maintainers and users?

pkosiec:
@simonfaltum The `updated_at` field isn't read or displayed anywhere in the CLI. We already display the skills version during their installation and I believe that's enough for the user 👍 

We have frequent conflicts on those fields and I'd suggest removing them if they are unused. Does it make sense?

simonfaltum (approved):

PR #73, opened 2026-05-12
experimental: import ai-dev-kit skills into experimental/ directory

jamesbroadhead:
## Summary

Adds an `experimental/` directory containing 18 agent skills from [databricks-solutions/ai-dev-kit](https://github.com/databricks-solutions/ai-dev-kit) `databricks-skills/`, imported as a snapshot on a **best-effort basis**. Excluded:
- `databricks-model-serving` (TODO #1b — different surface than stable, heavy MCP coupling)
- `databricks-spark-declarative-pipelines` (TODO #5 — different surface than stable `databricks-pipelines`)
- `databricks-genie` — removed during review per @lennartkats-db; deferred to a future revision
- `databricks-lakebase-provisioned` — not in the upstream `experimental` branch either

The manifest exposes both stable and experimental skills in a **single `skills` map**. Each entry carries a `repo_dir` field (`"skills"` or `"experimental"`) that points to the directory the skill lives in. Consumers derive experimental state from `repo_dir` — there is no parallel `experimental_skills` map and no per-skill `experimental` bool.

Paired with [databricks/cli#5243](https://github.com/databricks/cli/pull/5243) which teaches `databricks aitools install` (top-level) to:
- read `repo_dir` and skip experimental entries by default,
- install all of them with `--experimental`,
- install one by name with `--experimental` required.

Experimental and stable skills install under their plain names (e.g. `~/.claude/skills/databricks-iceberg/`); the upstream repo enforces name-uniqueness across both directories, so no install-side suffix is needed.

## Source

Final sync at merge time was [`20a92a3`](https://github.com/databricks-solutions/ai-dev-kit/commit/20a92a38144ca5228f1dfe4cc0be46da40ec9177) on `databricks-solutions/ai-dev-kit:experimental` ("tests: outcome-oriented rewrite across 7 skills + strip MCP tool_modules from all manifests").

Initial import was [`9c7a5b3`](https://github.com/databricks-solutions/ai-dev-kit/commit/9c7a5b3a3bf187c2b19d0b777768ecb52dd2de22) (head of [a-d-k PR #533](https://github.com/databricks-solutions/ai-dev-kit/pull/533) on the `appkit-on-experimental` branch). PR #533 has since merged into `experimental` (`7b07f18`). The branch went through periodic re-syncs during review to pull in upstream updates.

The rename (`databricks-app-python` → `databricks-apps-python`) is preserved in the merged version, which is what prevents a 3rd skill-name collision with d-a-s's stable `databricks-apps`.

[...]

PR #85, opened 2026-05-24
skills(pipelines): port databricks-spark-declarative-pipelines from a-d-k

jamesbroadhead:
## Summary

Ports the `databricks-spark-declarative-pipelines` skill from [`databricks-solutions/ai-dev-kit`](https://github.com/databricks-solutions/ai-dev-kit/tree/experimental/databricks-skills/databricks-spark-declarative-pipelines) into stable `skills/databricks-pipelines/`. Source: `databricks-solutions/ai-dev-kit:experimental`.

Completes d-a-s [PR #73](https://github.com/databricks/databricks-agent-skills/pull/73)'s TODO #5. Pairs with a-d-k [PR #546](https://github.com/databricks-solutions/ai-dev-kit/pull/546), which tombstones the a-d-k skill once this lands.

Stable's `databricks-pipelines` already covered the per-feature × per-language API/options surface (decision tree, common traps, format options, dataset/flow/quality references). a-d-k's version covered scaffolding/workflows, configuration, performance tuning, DLT migration, and several streaming patterns + Kafka ingestion + SCD-2 query patterns that stable lacked. This PR adds a-d-k's net-new content as new `references/` files; the per-feature reference structure is preserved.

### New `references/`

- `dlt-migration.md` — both migration paths (DLT Python → SDP Python via `pyspark.pipelines`, DLT Python → SDP SQL) with side-by-side conversions for decorators, reads, expectations, CDC/SCD, and partitioning → liquid clustering.
- `workflows.md` — Workflow A/B/C chooser (standalone bundle via `databricks pipelines init`, pipeline-in-existing-bundle, rapid CLI iteration with no bundle); language-selection rules; start-update + poll-the-update pattern (with the "never poll top-level pipeline state because RETRY_ON_FAILURE flips it back to RUNNING" rationale); edit/re-upload/restart flow; Python SDK alternative.
- `pipeline-configuration.md` — Full JSON config reference for `pipelines create|update` (top-level fields, `clusters`, `event_log`, `notifications`, `configuration`, `run_as`, `restart_window`, `environment`, `deployment`); variant snippets (dev mode, non-serverless, continuous, notifications, autoscaling, custom event log, serverless Python deps); multi-schema patterns; platform constraints.
- `performance.md` — Liquid Clustering with per-layer key guidance (bronze/silver/gold); cluster-key type rules; table properties; state-management strategies for streaming; join optimization (stream-to-static, stream-to-stream with time bounds); query optimization; pre-aggregation; compute config; monitoring.
[...]

PR #90, opened 2026-05-26
skills(execution-compute): drop bundled compute.py, document the CLI surface

jamesbroadhead:
## Summary

Per Lennart's [audit](https://github.com/databricks/databricks-agent-skills/pull/73#issuecomment-4534752055) on #73, item #7: *"for getting this out of experimental I think we need to just rely on the Databricks CLI. And if that somehow has gaps then we should fix them!"*

Doing this now rather than waiting for stable promotion — the CLI already covers everything `scripts/compute.py` wrapped, just verbosely for the cluster-code-execution path.

## Removed

`experimental/databricks-execution-compute/scripts/compute.py` (743 lines, three subcommands).

## Replacement mapping

| `compute.py` subcommand | CLI replacement |
|---|---|
| `list-compute --resource clusters` | `databricks clusters list --output json` |
| `list-compute --resource node-types` | `databricks clusters list-node-types` |
| `list-compute --resource spark-versions` | `databricks clusters spark-versions` |
| `list-compute --auto-select` | `databricks clusters list --cluster-sources UI,API --output json \| jq '.[] \| select(.state=="RUNNING")'` |
| `manage-cluster --action create/start/restart/terminate/delete` | `databricks clusters create/start/restart/delete/permanent-delete` |
| `execute-code --compute-type serverless` (upload+submit+wait+cleanup) | The existing `databricks workspace import` + `databricks jobs submit` + `databricks jobs get-run` + `databricks jobs get-run-output` flow that SKILL.md already documents. The "Convenience wrapper" section pointing at `compute.py` is removed. |
| `execute-code --compute-type cluster` (code on a running cluster with persisted context) | Multi-step `databricks api post /api/1.2/contexts/create` → `/api/1.2/commands/execute` → poll `/api/1.2/commands/status` → optional `/api/1.2/contexts/destroy`. Full recipe (create context, submit, poll, fetch results, follow-up commands reusing `contextId`, `%pip install`, language switching) added to `references/3-interactive-cluster.md`. |

## Other changes

- Removed the `<SKILL_ROOT>` path-convention note from SKILL.md (was only there to reference `scripts/compute.py` paths).
- "CLI Commands" table replaced with a more useful "CLI Command Map" covering upload, submit, run state, run output, cluster lifecycle, code-execution-on-cluster, and warehouses.
- Common-issues table for serverless updated to reference the `tasks[].timeout_seconds` field on submit JSON instead of `--timeout` on the wrapper.

## Test plan

- [x] `python3 scripts/skills.py generate` clean.
- [x] `python3 scripts/skills.py validate` passes.
- [ ] CI green.
- [ ] Reviewer sanity-check of the cluster-execution recipe in `references/3-interactive-cluster.md`.


experimental/databricks-execution-compute/SKILL.md
```diff
@@ -13,9 +13,7 @@ description: >-
 
 # Databricks Execution & Compute
 
-Run code on Databricks. Three execution modes—choose based on workload.
-
-> **Path convention:** `<SKILL_ROOT>` in examples below = the directory containing this SKILL.md. Resolve it to the absolute path in your install (e.g. `~/.claude/skills/databricks-execution-compute`). Commands like `python <SKILL_ROOT>/scripts/compute.py ...` work from any cwd.
+Run code on Databricks. Three execution modes—choose based on workload. All examples below use the Databricks CLI; install with `pip install databricks-cli` (or follow the workspace-native `databricks` quickstart) and authenticate with `databricks auth login` (see the parent `databricks-core` skill for profile setup).
```
lennartkats-db: This installation instruction is for the legacy CLI; we don't use pip anymore

experimental/databricks-execution-compute/references/2-serverless-job.md
```diff
@@ -109,7 +91,7 @@ Long dependency list from a file:
 | `ModuleNotFoundError` | Add the package to the environments spec with `"client": "4"` |
 | Dependencies listed but not installed | `"client": "1"` silently drops `dependencies`; use `"client": "4"` |
 | `get-run-output` returns empty `notebook_output` | You passed the parent run_id, not `.tasks[0].run_id` |
-| Job times out | Default 1800 s on the script wrapper; raise `--timeout` or use `jobs submit --no-wait` + your own polling |
+| Job times out | Use `databricks jobs submit --no-wait` and poll `get-run` yourself, or set `tasks[].timeout_seconds` in the submit JSON to extend the per-task limit |
```
lennartkats-db, line 94:
```
databricks api get … --json '{…}' for status polling is fishy. In references/3-interactive-cluster.md §3, the polling loop sends a JSON body on an HTTP GET to /api/1.2/commands/status.
  That endpoint takes clusterId/contextId/commandId as query string parameters, and HTTP GET bodies are usually dropped. The replacement should be:
  databricks api get "/api/1.2/commands/status?clusterId=$CID&contextId=$CTX&commandId=$CMD
```

lennartkats-db (changes requested):
Minor comments

experimental/databricks-execution-compute/SKILL.md
jamesbroadhead: (Claude here, on James's behalf.)

Fixed in 3d4d88b — dropped the `pip install databricks-cli` reference and now point readers at the `databricks-core` skill for install + auth.

Also did a wider grep — found one more carryover in `skills/databricks-pipelines/references/workflows.md` (came in from the a-d-k port in #85, originally `databricks-skills/databricks-spark-declarative-pipelines/references/1-project-initialization.md:418` upstream). Fixed in the same commit. There's also a stale error message in `ai-dev-kit`'s `databricks-mcp-server/databricks_mcp_server/tools/workspace.py:211` recommending the same legacy package — that one's out of scope here but worth flagging upstream.

experimental/databricks-execution-compute/references/2-serverless-job.md
jamesbroadhead: (Claude here.)

Confirmed your read and fixed in 7ebcdb0.

Looked at alternatives before patching: the modern Databricks CLI doesn't ship a `command-execution` subcommand (just checked v0.299.2 — no group covers the 1.2 endpoints), and the modern SDKs route the same way under the hood. So `databricks api get` with query-string params is the only correct surface for this legacy API. Done as you suggested:

```bash
databricks api get "/api/1.2/commands/status?clusterId=${CID}&contextId=${CTX}&commandId=${CMD}"
```

Added a one-liner above the snippet noting why the GET body has to go in the URL.

experimental/databricks-execution-compute/references/3-interactive-cluster.md
```diff
@@ -23,7 +21,7 @@
 **Starting a cluster takes 3-8 minutes and costs money.** Always check first:
 
 ```bash
-python <SKILL_ROOT>/scripts/compute.py list-compute --resource clusters
+databricks clusters list --output json | jq '.[] | select(.state == "RUNNING") | {cluster_id, cluster_name, state, cluster_source}'
```
QuentinAmbard: I believe we need to filter (server side) by default on 
--cluster-sources UI,API
Otherwise it's super slow in our env so we want to have this default in the example

QuentinAmbard (approved):

jamesbroadhead:
@lennartkats-db — (Claude on James's behalf.) Both your review comments are addressed:

- `pip install databricks-cli` → dropped; points at the `databricks-core` skill for install + auth (commit `3d4d88b`).
- GET-with-body on `/api/1.2/commands/status` → rewritten to use query-string params; the modern CLI has no `command-execution` subcommand so this is the canonical surface (commit `7ebcdb0`).

Replies on both threads. Quentin approved separately. When you have a minute, re-review so this can land.

experimental/databricks-execution-compute/references/3-interactive-cluster.md
jamesbroadhead: Done in 377b1c1. Also applied to line 163 which had the same issue — line 36 already had the filter so just bringing the other two in line with that convention.

simonfaltum (commented):
Reviewed the CLI surface this replaces and resolved the merge conflict. Looks good, with two small doc cleanups I pushed directly.

**Merge conflict:** only `manifest.json` (auto-generated). `main` had rewritten it (alphabetically sorted keys, version bumped to `0.1.0`) and moved skill references into `references/` subdirs. Resolved by regenerating with `python3 scripts/skills.py generate`; `validate` passes. The merged result matches this PR's intent exactly (+139/-865 across the same 6 files), and the regenerated `databricks-execution-compute` entry correctly drops `scripts/compute.py`.

**Verified against CLI v1.0.0** (since the recipes make concrete CLI claims):
- All `databricks clusters` subcommands exist as documented: `create` (takes `SPARK_VERSION` positionally), `delete` (= "Terminate cluster", reversible), `permanent-delete`, `restart`, `resize`, `start`, `get`, `list` (`--cluster-sources` flag), `list-node-types`, `spark-versions`.
- `databricks api get/post` exist; polling `commands/status` via query string (not a GET body) is correct.
- The `1.2 commands` API recipe is accurate end to end: `contexts/create` → `commands/execute` → `status` → `destroy`/`cancel`, the terminal-state set (`Finished`/`Error`/`Cancelled`), and the `.results.resultType` shapes (`text`/`error`/`table`).
- Dropping legacy `pip install databricks-cli` and switching to `tasks[].timeout_seconds` / `jobs submit --no-wait` are both correct.
- No leftover references to `compute.py`, `execute-code`, `list-compute`, `manage-cluster`, or `<SKILL_ROOT>` anywhere in the repo.

**Cleanups I pushed (2eda268), in `references/3-interactive-cluster.md`:**
- Removed a duplicate `clusters list` block in the "Never Start a Cluster Without Asking" section (kept the user-created-vs-job-cluster rationale inline under the first block).
- Made the recipe use a single `$CID` variable across create/execute/status/destroy instead of mixing a shell var with inline literals, so it copy-pastes as one sequence.

simonfaltum (approved):

atilafassina (approved):

lennartkats-db (approved):

PR #95, opened 2026-05-26
scripts(skills): drop dead base_revision plumbing, enforce canonical sorted manifest

jamesbroadhead:
## Summary

Two related cleanups to `manifest.json` hygiene.

### 1. Drop `base_revision` plumbing

The `base_revision` field was added in [#30](https://github.com/databricks/databricks-agent-skills/pull/30) as a per-skill upstream-revision tag, intended to track where a synced skill came from in a-d-k. It never got a consumer:

- `databricks/cli`'s `aitools install` ignores it (`gh search code base_revision --owner=databricks` returns only this repo's own `scripts/skills.py`).
- No skill on `main` carries it. A few feature branches populated it while iterating on a-d-k ports (e.g. `origin/experimental-aidevkit` had `"base_revision": "e742f36e8ab1"` for some skills), but those values never reached `main`.
- The generator just round-tripped the field if present in the prior manifest, and `normalize_manifest` stripped it before the validate diff — so it couldn't fail CI either.

It's dead weight that future maintainers have to reason about. Dropped:

- `_build_stable_entry`'s existing-entry-preservation branch (and the now-unused `existing_skills` parameter).
- `_build_experimental_entry`'s equivalent.
- `generate_manifest`'s existing-manifest read (only used to feed `existing_skills`).
- `normalize_manifest` and `_normalize_skill_map` — the only volatile field they normalized was `base_revision`, so `validate_manifest` can compare dicts directly now.

If per-skill upstream sync tracking does become useful later, re-add the field at the same time as the tool that consumes it. The long-term sync plan in [#73 TODO #3](https://github.com/databricks/databricks-agent-skills/pull/73) is git-subtree-based, which tracks revisions intrinsically — `base_revision` doesn't help there.

### 2. Enforce canonical sorted form

The on-disk `manifest.json` is now required to be byte-equal to the canonical serialization (`json.dumps(manifest, indent=2, sort_keys=True) + "\n"`). That means:

- The `skills` map's keys are alphabetical across stable **+** experimental (no more stable-first grouping; that was a side-effect of build order, not a deliberate choice).
- Each skill entry's keys are alphabetical: `description`, `files`, `min_cli_version?`, `repo_dir`, `version`.
- The top-level keys (`skills`, `version`) are alphabetical.
- `files` arrays remain sorted by the generator (already enforced).

`scripts/skills.py validate` now does two checks:

1. **Content lint** — the parsed dict equals what `generate_manifest` would produce. Catches stale content, unsorted/missing/extra `files` entries, drift between SKILL.md frontmatter and manifest.
2. **Canonical-form lint** — the on-disk bytes equal `serialize_manifest(current)`. Catches hand-edited sort drift (re-ordered skill names, re-ordered per-skill keys) even when the parsed content is correct.

The writer at `scripts/skills.py generate` uses the same `serialize_manifest` helper, so files emitted by the generator always pass the lint.

## Diff shape

| File | Change |
|---|---|
| `scripts/skills.py` | −47 / +35 lines: drop `base_revision` branches + dead normalize machinery; add `serialize_manifest` helper; rework `validate_manifest` to do the two checks above. |
| `manifest.json` | 220-line swap: skills now in one alphabetical run (`databricks-agent-bricks`, `databricks-ai-functions`, ...) and per-skill keys alphabetised. No semantic change for `aitools install`. |

## Test plan

- [x] `python3 scripts/skills.py generate` clean — regenerates the canonical sorted manifest.
- [x] `python3 scripts/skills.py validate` passes.
- [x] Lint catches a manually-reversed `skills` map: emits `ERROR: manifest.json is not in canonical sorted form. Keys must be alphabetical at every level.`
- [x] Lint catches a manually-reversed `files` array inside a skill entry: emits `ERROR: manifest.json content is out of date`.
- [ ] CI green on this branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,19 +95,21 @@ This approach:
 
 ### Manifest Management
 
-Sync assets and generate manifest after adding/updating skills:
+`manifest.json` is **generated** by `scripts/skills.py` from the skill directories and frontmatter. Do not edit it by hand. CI rejects manual changes via two checks: content drift (parsed dict doesn't match what `generate` would produce) and canonical form (on-disk bytes don't match `json.dumps(..., indent=2, sort_keys=True)`).
+
+Sync assets and regenerate the manifest after adding or updating skills:
 
 ```bash
 python3 scripts/skills.py
 ```
 
-Validate that assets and manifest are up to date (for CI):
+Validate that assets and manifest are up to date (used by CI):
 
 ```bash
 python3 scripts/skills.py validate
 ```
 
-The manifest is used by the CLI to discover available skills.
+The manifest is consumed by the CLI to discover available skills.
 
 ## Security
 
--- a/manifest.json
+++ b/manifest.json
@@ -1,4 +1,5 @@
 {
+  "//": "GENERATED FILE: DO NOT EDIT. Run `python3 scripts/skills.py generate` to regenerate.",
   "skills": {
     "databricks-agent-bricks": {
       "description": "Create Agent Bricks: Knowledge Assistants (KA) for document Q&A and Supervisor Agents for multi-agent orchestration (MAS).",
--- a/manifest.json
+++ b/manifest.json
@@ -1,10 +1,51 @@
 {
-  "version": "2",
   "skills": {
+    "databricks-agent-bricks": {
+      "description": "Create Agent Bricks: Knowledge Assistants (KA) for document Q&A and Supervisor Agents for multi-agent orchestration (MAS).",
+      "files": [
+        "SKILL.md",
+        "agents/openai.yaml",
+        "assets/databricks.png",
+        "assets/databricks.svg",
+        "references/1-knowledge-assistants.md",
+        "references/2-supervisor-agents.md"
+      ],
+      "repo_dir": "experimental",
+      "version": "0.0.1"
+    },
+    "databricks-ai-functions": {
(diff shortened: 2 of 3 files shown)


simonfaltum:
**Follow-up idea: drop `SKILL_METADATA`**

`SKILL_METADATA` at `scripts/skills.py:23` carries one field per stable skill: `description`. Each skill already has a `description:` in its `SKILL.md` frontmatter, and `_build_experimental_entry` reads from frontmatter via `extract_description_from_skill`. Stable skills duplicate, and the dict copy wins silently.

The two strings have already drifted on `main`. For example, `databricks-apps`:
- `SKILL_METADATA`: "Databricks Apps development and deployment (evaluates analytics vs synced tables data access)"
- `SKILL.md` frontmatter: "Build apps on Databricks Apps platform. Use when asked to create dashboards, data apps, analytics tools, or visualizations. Invoke BEFORE starting implementation."

Reading the two together, it looks like they serve different audiences: the dict is a terser blurb for `aitools list` / marketplace, and the frontmatter carries "Use when..." trigger phrases for the agent. Worth deciding on purpose, then collapsing:

**Option A: frontmatter is the source of truth.** `_build_stable_entry` reads description from frontmatter, same path as experimental. Use the existing `synthesize_short_description()` to strip "Use when..." for the marketplace blurb if we want a clean version in `aitools list`. Single source, no drift.

**Option B: keep the dict but enforce sync.** Add a validate check that the dict and frontmatter agree (or that one is a prefix of the other). Catches drift in CI but keeps the two-field model.

**Option C:** If we leave it as is, we will have two intentionally different descriptions and accept the drift.

I lean Option A. The frontmatter is already authoritative for experimental skills, the strip helper already exists, and the current drift is silent rather than deliberate.

Not blocking this PR, just flagging as a follow-up. Thoughts?

simonfaltum (approved):
Approving the base PR (drop base_revision + canonical sorted form). Pushed a small follow-on commit with a generated-file banner and clarified README wording. Left a separate comment on the SKILL_METADATA duplication as a follow-up idea.

PR #108, opened 2026-05-27
release: switch workflow to annotated tags + add RELEASING.md

jamesbroadhead:
## Summary

The release workflow used `gh release create "$VERSION"` which creates a **lightweight** tag (`objecttype=commit`). `git tag -v` cannot verify lightweight tags — they have no tag object to attach a signature to. `v0.1.0` through `v0.2.1` are all lightweight, so the README "Integrity" claim that future tags are GPG-signed and verifiable couldn't be honoured even if signing were turned on.

This PR makes annotated tags the default in the release workflow — the strict prerequisite for `git tag -v` and for GPG/Sigstore signing.

## Changes

- **`.github/workflows/release.yml`** — split tag creation from release creation. `git tag -a "$VERSION" -m "$VERSION"`, push the tag, then `gh release create --verify-tag` (confirms the tag exists before the release).
- **`README.md` Integrity** — stop claiming "future tags will be GPG-signed" as a present fact; describe the actual state (annotated now, signed pending GPG/Sigstore enablement) and link to RELEASING.md.
- **`RELEASING.md`** — new doc covering the release procedure, `git tag -v` verification, and the signing follow-up.

## Out of scope (follow-up)

Signing itself. It requires either:

- **GPG**: release-identity key + private key/passphrase in GH Actions secrets; sign step uses `git tag -s`.
- **Sigstore (gitsign)**: install `sigstore/gitsign` in the workflow; no long-lived secret; OIDC token is the identity.

Either approach satisfies the README claim. The choice is a team decision; annotated tags are the strict prerequisite that this PR unblocks.

Existing `v0.1.x` / `v0.2.x` tags stay lightweight (would need re-tagging to convert) — RELEASING.md covers both options for that question.

## Test plan

- [x] `python3 scripts/skills.py validate` passes (no skill content touched).
- [ ] Reviewer to dry-run the workflow on a throwaway tag (`v0.0.0-rc1`) in a follow-up to confirm annotated-tag creation works on the protected runner.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,11 +27,24 @@ jobs:
             exit 1
           fi
 
-      - name: Create tag and release
+      - name: Create annotated tag
+        env:
+          VERSION: ${{ inputs.version }}
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git tag -a "$VERSION" -m "$VERSION"
+          # Tag is annotated (objecttype=tag) so `git tag -v <tag>` can verify
+          # signatures once GPG/Sigstore signing lands. Today the tag is
+          # annotated but unsigned — see RELEASING.md.
+          git push origin "$VERSION"
+
+      - name: Create release
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           VERSION: ${{ inputs.version }}
         run: |
           gh release create "$VERSION" \
             --title "$VERSION" \
-            --generate-notes
+            --generate-notes \
+            --verify-tag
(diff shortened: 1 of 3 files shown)


simonfaltum:
Thanks for this @jamesbroadhead. Since you opened it, #111 landed and touched the same two files: it already softened the README claim and added a Releasing section to `CONTRIBUTING.md`, and it restructured `release.yml` (the version-bump + commit steps), which is what's causing the conflicts here.

I pulled the still-relevant part (the annotated-tag switch) into #116, rebased on current main so the tag is created after the version-bump commit. I also dropped the GPG-signing line from the README entirely rather than rewording it, since we're not committing to signing right now. I left `RELEASING.md` out because `CONTRIBUTING.md` already documents the release flow.

Suggest we close this in favor of #116. Thanks again for catching the lightweight-tag issue.

PR #115, opened 2026-06-01
Add Scala JAR migration to databricks-serverless-migration

nachill-databricks:
## What
Extends the `databricks-serverless-migration` skill to handle compiled Scala JARs (`spark_jar_task`). Today the skill covers Python/SQL/notebooks deeply, but for Scala it only names the requirements ("compile as JAR, Scala 2.13, JDK 17, env 4+") without guiding the actual build/dependency migration. This adds that.

## Changes
- **`references/jar-migration.md`** (new):
  - Four failure modes with exact signatures (Scala 2.12→2.13 `wrapRefArray`, Spark internals, dependency shadowing, streaming/config).
  - sbt fixes: Scala 2.13.16, `databricks-connect % Provided` (the artifact serverless actually provides), conflicting deps `% Provided`, exclude Scala from the assembly, preserve `META-INF/maven`.
  - The env-4 kernel classpath table (Scala 2.13.16 / JDK 17 / Databricks Connect 17.3.1) so conflicts can be flagged statically — matching the runtime's Dependency Conflict Detection output.
  - Transitive dependency scan (`sbt dependencyTree`), build/test/deploy gates, and a production-rigor path (test branch + sampled data + A/B comparison + PR) mirroring the parent skill's flow.
- **`SKILL.md`**: decision-tree branch for JAR workloads, a Category E row, and a reference-index entry.
- **`manifest.json`**: regenerated via `scripts/skills.py generate` (`validate` passes).

## Scope
- sbt-first (the common case); Maven/Gradle equivalents noted, full recipes TODO.
- The classpath table is the env-4 snapshot; the runtime diagnostic remains the live source of truth.

## Testing
Manually validated end to end against a real classic-compute Scala JAR job (`spark_jar_task`). I applied this skill's procedure to migrate it to serverless — recompiled from Scala 2.12 to 2.13.16, marked `databricks-connect` and the conflicting bundled deps `% Provided`, and moved the job to environment version 4. The migrated job deployed and ran successfully on serverless (`TERMINATED / SUCCESS`). The two failure modes the JAR exercised (Scala version mismatch and dependency shadowing) matched the reference's signatures and were resolved by the documented fixes.

### DIFF
--- a/skills/databricks-serverless-migration/references/jar-migration.md
+++ b/skills/databricks-serverless-migration/references/jar-migration.md
@@ -1,0 +1,206 @@
+# Scala JAR Migration Reference
+
+Deep reference for migrating a **compiled Scala JAR** (`SPARK_JAR_TASK`) from classic compute to serverless. The parent `databricks-serverless-migration` skill routes here when the workload is a JAR rather than a notebook or script.
+
+Serverless runs JARs on a Spark Connect kernel with a **fixed classpath**: **Scala 2.13.16, JDK 17, Databricks Connect 17.3.1** (environment version 4). Migration is about making the JAR match that classpath instead of bundling its own conflicting copies.
+
+## When the parent skill delegates here
+
+Route into this flow when any of these are true:
+- A `spark_jar_task` / `SPARK_JAR_TASK` job, or a build that produces a JAR (`build.sbt`, `pom.xml`, `build.gradle`).
+- A serverless run whose output contains a `>>> Scala Version Check` or `>>> Dependency Conflict Detection` block.
+- A `NoSuchMethodError: scala.Predef$.wrapRefArray`, `NoClassDefFoundError: scala/Serializable`, or `NoClassDefFoundError` for a Spark-internal class on a serverless JAR run.
+
+## The four failure modes (cover 100% of classified production JAR failures)
+
+| # | Failure | Signature | Root cause | Fix |
+|---|---|---|---|---|
+| 1 | **Scala version mismatch** | `NoSuchMethodError: scala.Predef$.wrapRefArray`, `NoClassDefFoundError: scala/Serializable` | JAR compiled against Scala 2.12 (classic default). Serverless is 2.13.16. Binary-incompatible at classload. | Recompile against **2.13.16**; `%%` cross-version on every Scala dep. |
+| 2 | **Spark internals not on classpath** | `NoClassDefFoundError` for `org/apache/spark/...` driver-side classes (`JavaSparkContext`, Catalyst internals) | Code reaches behind the Spark Connect boundary; Spark is provided by the runtime, not bundled. | Mark Spark `% Provided` (don't bundle it). If the **source** uses `SparkContext`/RDD/Catalyst, rewrite to the DataFrame / Spark Connect surface — a code change, not just a build change. |
+| 3 | **Dependency conflict / shadowing** | Jackson/Guava/log4j version errors; behavior differs from the bundled version | The JAR bundles a library the kernel also ships; the kernel's copy wins on the classpath, so the bundled version is silently shadowed. | Declare every overlapping dependency `% Provided` (or align to the kernel version). See the classpath table below. |
+| 4 | **Streaming / config** | `ProcessingTime` trigger errors, blocked Spark configs | Serverless requires `availableNow` triggers and only allows a short config allowlist. | Source change: `.trigger(availableNow=true)`; remove unsupported `spark.conf.set`. |
+
+## Step 1 — Analyze the build statically (do not wait for a failed run)
+
+Read the build file and flag issues before running anything.
+
+**Scala version.** In `build.sbt`, find `scalaVersion`. If it is not `2.13.x`, that is failure mode 1.
+```
+grep -nE 'scalaVersion\s*:=' build.sbt
+```
+Also flag any explicit `_2.12` artifact suffix and any Scala dep declared with single `%` (no cross-version) instead of `%%`.
+
+**JDK version.** Serverless runs **JDK 17**. A JAR compiled with a newer JDK (21+) fails on the kernel. Check the build's `javacOptions`/`-release` and the local `java -version`, and compile with JDK 17. (JDK 8/11 bytecode runs fine — forward-compatible.)
+
+**Bundled Spark.** Any `spark-core`, `spark-sql`, `spark-catalyst` dependency that is *not* marked `% Provided` is bundling Spark (mode 3). The fix is to mark it `% Provided` (the runtime supplies it). **Watch for mode 2 separately:** if the *source* uses `SparkContext`, `sc.parallelize`, RDD APIs, or Catalyst internals, marking Spark provided is not enough — that code must be rewritten to the DataFrame/Spark Connect surface (a source change, not just a build change). Flag it so the build gate expects a recompile.
+
+**Conflicting dependencies — scan the *full* tree, not just direct deps.** Most conflicts are transitive: our demo JAR pulls guava, log4j, json4s, and Jackson *through* `spark-sql`, never declaring them directly. Matching only `libraryDependencies` misses them. Enumerate what the JAR will actually bundle and match every entry against the kernel classpath table:
+```
+sbt 'set asciiGraphWidth := 240' dependencyTree   # full transitive tree
+sbt evicted                                        # version conflicts already resolved by sbt
+# Maven:  mvn dependency:tree
+# Gradle: ./gradlew dependencies --configuration runtimeClasspath
+```
+Every node that appears in the classpath table is a conflict (mode 3). Fix the **whole set** at once: mark the direct deps `% Provided`, and for transitive offenders pulled by another dep, either mark that parent `% Provided` (removes the whole subtree) or add an explicit `% Provided` override. Confirm with `dependencyTree` again that no table entry remains in the assembled set. This is also what lets the skill reproduce the runtime's `Dependency Conflict Detection` list before any run.
+
+**Assembly metadata.** If `assemblyMergeStrategy` discards `META-INF/maven/**`, the runtime cannot read exact bundled versions (it falls back to class-name matching, and says so in the diagnostic). Recommend keeping it:
+```scala
+case PathList("META-INF", "maven", _ @ _*) => MergeStrategy.first
+```
+
+## Step 2 — Apply the fixes (sbt)
+
+**Scala version → 2.13.16:**
+```diff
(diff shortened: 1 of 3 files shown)


rrf-DB (approved):
LGTM with the suggestions below as non-blocking follow-ups. Please merge after databricks/databricks-agent-skills#72 lands so the two SKILL.md edits don't fight; I tested the combined tree locally and the conflict is mechanical (decision-tree, pattern-table row, references list — clean three-way merge).

**Regression** (run against this PR rebased on top of #72, against our internal test bench)
- `lint`: PASS — all references valid markdown, code blocks parse
- `isa`: 16/16 SUCCESS (`isa-cv-quality-inspection` timed out on first attempt and succeeded on retry in 7 min — a known infra-flaky case that's been borderline at the 15-min timeout for weeks across the same set of runs; doc-only PR can't have caused it)
- `compat`: PASS — Prashanth's 15-notebook compatibility suite
- `sanity`: PASS — SKILL.md resolves all 10 references (was 9, +1 for the new `jar-migration.md`)

**Strengths**
- Concrete failure modes with exact error signatures (search-discoverable).
- Real version numbers (Scala 2.13.16 / JDK 17 / Databricks Connect 17.3.1 / env 4) with an explicit "env 5 may differ" caveat. Right level of calibration.
- Mode 2 is correctly framed as a *source* change, not a build change (catches the common "mark Spark Provided and call it done" mistake).
- Cheapest-gate-first ordering: build → local Databricks Connect → deploy. Saves a deploy on a JAR that didn't assemble.
- Production-rigor path defers to the parent skill's two-branch + A/B + PR flow instead of inventing a parallel methodology. Single JAR-specific addition (parameterizing hardcoded catalog/schema) is the right minimal delta.
- DABs YAML in section 4 nails the "rejected at deploy time, not run time" gotcha: `environment_key` not `job_cluster_key`, `java_dependencies` not task-level `libraries`, env_version "4".
- Narration-at-every-step style matches the parent skill's voice.

**Suggestions (non-blocking, consider in a follow-up)**

1. **Build-system invocation gate** — Step 1 + Step 3 call `sbt dependencyTree` / `mvn dependency:tree` / `gradle dependencies`, which execute the workload's build system. The parent skill's `multi-source-enumeration.md` recently added an AST-only-parse + user-confirmation gate for the symmetric concern (per a recent security review on #72). Either add a one-line precondition ("These commands run the workload's build; confirm with the user that the build is trusted before invoking") or note that this is acceptable because the user is migrating their *own* workload.
2. **Runtime diagnostic strings** — "a serverless run whose output contains a `>>> Scala Version Check` or `>>> Dependency Conflict Detection` block." Worth a one-line citation (doc or runtime change). Otherwise readers may search for strings the kernel doesn't emit yet.
3. **"100% of classified production JAR failures"** — strong claim; consider "the four failure modes we've seen across migrations to date" if the data source can't be cited publicly. Reads as marketing in an otherwise technical doc.
4. **`assemblyMergeStrategy`** snippet — `MergeStrategy.first` for `META-INF/maven/**` is correct; consider noting that a project with an existing `assemblyMergeStrategy` should *add* this branch rather than replace.
5. **Step 3.2 `DatabricksSession.builder().serverless()` bootstrap** — the note "add one if migrating from a bare `SparkSession.builder()`" is good; a one-line example would save the reader a context switch to `default-scala`.
6. **`environment_version: "4"`** — value is string-quoted ("4" not 4). Worth a copy-paste-protection note since the API will reject the int form.
7. **Maven/Gradle full recipes (TODO)** — fine for v1; consider opening a follow-up issue so it doesn't get lost.

Thanks for the contribution. The classpath collision table alone is going to save people hours.

This review was written by Isaac.

skills/databricks-serverless-migration/references/jar-migration.md
```diff
@@ -0,0 +1,254 @@
...
+| 2 | **Spark internals not on classpath** | `NoClassDefFoundError` for `org/apache/spark/...` driver-side classes (`JavaSparkContext`, Catalyst internals) | Code reaches behind the Spark Connect boundary; Spark is provided by the runtime, not bundled. | Mark Spark `% Provided` (don't bundle it). If the **source** uses `SparkContext`/RDD/Catalyst, rewrite to the DataFrame / Spark Connect surface — a code change, not just a build change. |
+| 3 | **Dependency conflict / shadowing** | Jackson/Guava/log4j version errors; behavior differs from the bundled version | The JAR bundles a library the kernel also ships; the kernel's copy wins on the classpath, so the bundled version is silently shadowed. | Declare every overlapping dependency `% Provided` (or align to the kernel version). See the classpath table below. |
+| 4 | **Streaming / config** | `ProcessingTime` trigger errors, blocked Spark configs | Serverless requires `availableNow` triggers and only allows a short config allowlist. | Source change: `.trigger(availableNow=true)`; remove unsupported `spark.conf.set`. |
```
simonfaltum: Tiny thing, since this reference is Scala-only: `.trigger(availableNow=true)` is the Python keyword-arg form, so it wouldn't compile if an agent pastes it into Scala source. Could we use the Scala equivalent here?

`.trigger(Trigger.AvailableNow())`  (with `import org.apache.spark.sql.streaming.Trigger`)

The rest of the row reads well.

skills/databricks-serverless-migration/references/jar-migration.md
```diff
@@ -0,0 +1,254 @@
...
+
+**Conflicting dependencies — scan the *full* tree, not just direct deps.** Most conflicts are transitive: our demo JAR pulls guava, log4j, json4s, and Jackson *through* `spark-sql`, never declaring them directly. Matching only `libraryDependencies` misses them. Enumerate what the JAR will actually bundle and match every entry against the kernel classpath table. (These commands invoke the workload's own build system — `sbt`/`mvn`/`gradle` execute project plugins and settings — which is expected here because the user is migrating their own JAR. If the build's provenance is unknown or untrusted, confirm with the user before invoking, mirroring the parent skill's `multi-source-enumeration.md` gate.)
```
simonfaltum: Small one: I went looking for `multi-source-enumeration.md` and couldn't find it in the parent skill (or anywhere in the repo today). The instruction reads clearly on its own, so perhaps we could just drop the "mirroring the parent skill's `multi-source-enumeration.md` gate" clause? If you'd rather keep a cross-reference, we could point at something that exists in the parent, like its Step 1 (Ingest) or Stopping Conditions. Either is totally fine.

simonfaltum (commented):
Thanks for this, the JAR path is a really nice addition and the job-config section lines up exactly with the Jobs API: `java_dependencies`, `environment_version` as a quoted string, and `dependencies` for pip all match the Go SDK. `scripts/skills.py validate` passes, and the CLI bits check out too (`default-scala` is a real `bundle init` template, and `dbfs:/Volumes/...` is the right scheme for `fs cp`).

Just two small one-liners inline that would be worth tidying before merge. Neither is a blocker.

One optional follow-up (not for this PR): the new section correctly uses `environment_version`, which is the current field. The Go SDK marks the older `client` field as deprecated ("Use environment_version instead"), and the parent skill still uses `client` in its Python/notebook examples, so it might be nice to modernize those for consistency in a separate PR.

Nice work overall.

simonfaltum:
A few follow-up suggestions, ranked by impact. None are blockers, the reference is solid as-is and the move to delegate testing back to the parent's Step 3 is the right call.

**1. Make the reference reachable when an agent is reacting to an error (highest impact).**
The decision tree routes here when an agent proactively scans a workload and spots a `spark_jar_task` or `build.sbt`. But the common entry point is reactive: a user pastes a failed serverless run and asks to fix it. That agent goes straight to the "Debug failed serverless runs" table in `SKILL.md`, and the Scala JAR signatures are not in it, so the whole reference never gets reached on that path. Suggest lifting the signatures the reference already documents into that table, each pointing here:

| Error | Fix |
|---|---|
| `NoSuchMethodError: scala.Predef$.wrapRefArray` / `NoClassDefFoundError: scala/Serializable` | Scala version mismatch on a JAR. Recompile against 2.13.16. See `references/jar-migration.md` |
| `NoClassDefFoundError` for `org/apache/spark/...` on a JAR run | Spark bundled instead of provided. See `references/jar-migration.md` |

Small and additive, and it is what makes the reference fire on the path users actually hit.

**2. Maven and Gradle are half-supported.**
The build-tool section ends with "Full Maven/Gradle fix recipes are a TODO, for now translate the sbt steps." A Maven user (common in enterprise Scala shops) follows the skill, hits a one-line gesture instead of a recipe, and stalls. The three moves are already stated (2.13 artifacts, scope conflicting deps `provided`, preserve `META-INF/maven`), so either add a compact `pom.xml` / `build.gradle` block mirroring the sbt one, or narrow the scope honestly in the decision tree and Category E row (this path covers sbt today). Either is fine. Leaving the TODO is the one option to avoid.

**3. Lead with the bundle deploy, and link the skills that own it.**
The fast path leads with `databricks fs cp` + `jobs run-now`, with bundles as an afterthought. Two sibling skills already own this surface, `databricks-dabs` and `databricks-jobs`, and neither is cross-linked. The bundle is the version-controlled end state the rest of this skill already pushes toward (the git-flow section puts `build.sbt` on a branch and opens a PR). Suggest reordering so `bundle deploy` is the durable default and `fs cp` + `run-now` is the quick inner-loop check, and pointing at those two skills for the bundle mechanics instead of re-teaching them. This mirrors what the reference already does well for testing.

**4. Routing: the description does not mention JARs or Scala.**
The skill's description says "notebooks and jobs" but never "Scala", "JAR", or "spark_jar_task", so a user asking to migrate a Scala JAR may not trigger the skill at all. The description is already on the long side, so swap rather than grow: fold "Scala JARs (`spark_jar_task`)" into the existing clause and trim something redundant. Worth a quick routing check against a few trigger phrases before adopting.

**Smaller things:**
- The parent teaches `spec.client` (`"1"`/`"2"`/`"3"`) for environment version, this reference uses `spec.environment_version: "4"`. An agent cannot tell if these are the same field, two fields, or one stale. Worth picking one story and stating it.
- The decision-tree line spells out "jackson, guava, log4j, slf4j, ..." which duplicates the classpath table here and will drift from it. Could collapse to "bundles libs on the kernel classpath? see the table in `references/jar-migration.md`".
- The "Output the skill should produce" checklist overlaps the parent's "Migration Deliverables" table. Could reference that and add only the JAR-specific rows (the `build.sbt` diff, the assemble result).

simonfaltum (approved):
Approving it to unblock, please take comments into consideration.

nachill-databricks:
isaac-bot pull master and resolve any conflicts.

PR #125, opened 2026-06-04
Add data-app-design skill (analytics/BI/AI app design)

keugenek:
## What

Adds **`data-app-design`** — a skill for designing the UX of Databricks analytics/BI/AI data apps (dashboards, KPI/overview pages, reports, charts, and **Genie/chat** surfaces) and mapping the design to concrete AppKit components. Also adds a routing pointer in **`databricks-apps`** so the entry skill invokes it for any data-displaying app.

## Provenance

Merges two skills authored by **Andre Furlan** in `andre-furlan_data/apps-usage` — `dashboard-design-patterns` (Bach et al., *Dashboard Design Patterns*, IEEE TVCG 2023) and `ibcs-reporting-standards` (IBCS SUCCESS) — into one AppKit-bound skill. Changes vs. the originals:
- Bind every pattern to real AppKit components/hooks/tokens (`references/appkit-cheatsheet.md`).
- Add a required-states/data-realism checklist and an AI-result-trust section for Genie.
- Fix the IBCS forecast/previous-year notation; tighten triggers + add when-NOT-to-use.
- Split catalogs into `references/` for progressive disclosure. `parent: databricks-apps`.

## Files
- `skills/data-app-design/**` — new skill + references
- `skills/databricks-apps/SKILL.md` — routing pointer to `data-app-design` for data UIs
- `manifest.json` — regenerated

## Eval (app-evals, dev-googfood)

- **Invocation reliable:** the generating agent invokes the skill **5/5** on dashboard prompts (incl. Genie) after the routing fix (was 0/5 — it only loaded `databricks-apps`).
- **No functional regression:** `appeval_100` = 1.00 on both arms; the skill arm used **fewer** generation turns (43.4 vs 53.4).
- **Design lift:** blind judge (3 perspective-diverse judges + adversarial verifier per app, anonymized & alternated), WITH vs WITHOUT on 5 dashboard prompts → **WITH 3/5 wins, mean 79.4 vs 73.8 (Δ +5.6)**. Wins show semantic tokens (dark-mode-safe), real AppKit `Table`/`Badge`/charts, and loading/empty/error states vs baselines that hardcoded hex and hand-rolled tables.

## Known follow-up
- **Harden the Genie/AI-trust section.** In the A/B the skill *lost* the Genie app — its AI-trust guidance reads as prose, not patterns. Convert to concrete copy-pasteable patterns (identity via `/api/whoami`, generated-SQL display, disclaimers).

Full report + reproduce commands: `apps-usage/AB_EVAL_REPORT.md`.


skills/databricks-app-design/references/genie-ai-trust.md
keugenek, line 1: this caused minor regression, rechecking

skills/databricks-app-design/references/appkit-cheatsheet.md
keugenek, line 1: @MarioCadenas should this be AppKit part?

keugenek:
Update: hardened the Genie/AI-trust section into concrete copy-pasteable patterns (`references/genie-ai-trust.md`: /api/whoami identity+OBO, Generated-SQL inspection Card, streaming/status, disclaimer, governance/empty states; commit `d50afc7`).

Re-ran the A/B on the Genie app — the earlier loss **flipped to a high-confidence 3-0 win** (WITH-hardened 86 vs baseline 78), judges crediting the generated-SQL review Card, disclaimer, and OBO/governance card. Dashboards unchanged (no regression). Net: reliably invoked (5/5), +design lift across the domain, no functional regression. Ready to merge.

(Written by Isaac.)

skills/data-app-design/references/appkit-cheatsheet.md
```diff
@@ -0,0 +1,79 @@
+# AppKit binding cheat-sheet
+
+The design guidance in this skill must compile to real code. Use these exact primitives — do
+not hand-roll charts, color, or state handling that AppKit already provides. Names verified
+against `@databricks/appkit-ui` and `@databricks/appkit` (v0.24) and this repo.
```
MarioCadenas: this shouldn't have a hardcoded version of appkit here no?

keugenek:
Addressed the review (commit `cf60363`):

- **@MarioCadenas — hardcoded appkit version:** dropped it. Slimmed `appkit-cheatsheet.md` to a design→primitive map that points at `@databricks/appkit-ui`/`@databricks/appkit` `CLAUDE.md` for the exact API (source of truth, won't drift).
- **"should this be AppKit part?":** agreed — the API surface now lives in AppKit's own docs; the skill keeps only the design→component/token mapping + pointers, not duplicated API listings.
- **"genie-ai-trust caused minor regression":** investigated — **no** Genie leakage into the dashboard apps (greps for whoami/useGenieChat/Generated-SQL/disclaimer are clean across all 3 generations). The cb_brickhouse 77→73 move was resampling noise (the newer generation is arguably richer; one app used raw `amber-*` instead of `var(--warning)` — a single-sample artifact, not from the Genie ref). Added a **scoping gate** so the Genie section + reference apply ONLY to apps with a Genie/chat/NL surface (skip for pure dashboards) — correct-by-construction.

(Written by Isaac.)

skills/databricks-app-design/references/genie-ai-trust.md
keugenek: Rechecked — no real regression. No Genie leakage into non-Genie apps (clean greps across all generated versions); the cb_brickhouse 77→73 design-judge move was resampling noise (the post-hardening generation is arguably richer). Shipped a scoping gate (cf60363) so this section + reference apply ONLY to Genie/chat/NL apps, never pure dashboards, and hardened the semantic-color rule to forbid raw Tailwind palette classes (8cec34c) — which covers the lone real nit (a dashboard using bg-amber-* instead of var(--warning)). Net: the Genie app flipped to an 86-vs-78 win after hardening. Resolving. (Written by Isaac.)

skills/databricks-app-design/references/appkit-cheatsheet.md
keugenek: Agreed. The cheat-sheet no longer embeds the AppKit API surface — it now points at the appkit-ui / appkit CLAUDE.md docs as the source of truth and keeps only the design→component/token mapping (cf60363). Resolving. (Written by Isaac.)

skills/data-app-design/references/appkit-cheatsheet.md
keugenek: Done — dropped the hardcoded version and slimmed the file to point at AppKit's own docs as the API source of truth (cf60363). Resolving. (Written by Isaac.)

keugenek:
@lennartkats-db @MarioCadenas — ready for your review/approval 🙏

All three review threads are resolved, CI (`validate`) is green, and the latest commits add the Genie scoping gate and harden the semantic-color rule (no raw hex **or** raw Tailwind palette classes). @lennartkats-db your approval is required for the net-new skill under `/skills/`; @MarioCadenas the cheat-sheet now drops the pinned version and points at the appkit-ui/appkit `CLAUDE.md` as the API source of truth, per your comment.

Summary + eval results (5/5 invocation, no functional regression, +5.6 design lift, Genie app 86-vs-78 win) are in the PR description. Thanks!

(Written by Isaac.)

simonfaltum:
Really nice work on this one, especially the eval setup.

One question I keep coming back to (not specific to this PR, but this PR leans into it): what decides whether the user gets a **Databricks App** or an **AI/BI dashboard** when they prompt "build me a dashboard"?

Today the routing effectively always answers "App": `databricks-apps` says "Use when asked to create dashboards" in its description, this PR adds a stable skill whose description leads with dashboards plus a MUST-use pointer for any data-displaying app, and `databricks-aibi-dashboards` is experimental, so it isn't in default installs (CLI stable set or the plugin marketplace). The two products differ a lot in price and maintenance (app compute plus code the customer owns and redeploys, vs a managed serverless dashboard with governance and sharing built in), and nothing in the flow surfaces that tradeoff or asks the user.

I'm wondering if we should:

- add a small carve-out to `data-app-design`'s description / when-NOT list, i.e. "not for authoring AI/BI (Lakeview) dashboards"
- longer term, add a product decision gate in `databricks-apps`, similar to the existing Data Access Decision Gate: show the tradeoff, let the user choose

Not blocking this PR on it, the routing question is bigger than this skill. But I think we need a deliberate answer before more dashboard-shaped polish lands on the Apps path. thoughts?

(Written by Isaac.)

skills/data-app-design/references/genie-ai-trust.md
```diff
@@ -0,0 +1,67 @@
...
+a spinner" is not enough. Snippets use the real `@databricks/appkit` + `@databricks/appkit-ui` APIs.
+
+## 1. Authenticated identity (who is asking — on-behalf-of)
+Databricks injects identity headers; expose them and show the caller. Queries run as that user (OBO).
+```ts
+// server/server.ts
+app.get("/api/whoami", (req, res) => {
+  res.json({
+    email: req.header("x-forwarded-email") ?? req.header("x-forwarded-user") ?? null,
+    userId: req.header("x-databricks-user-id") ?? null,
```
simonfaltum: **[important]** `x-databricks-user-id` doesn't look like a real platform header. The repo's own references use `x-forwarded-user` for the user id (apps `references/appkit/files.md`), plus `x-forwarded-email` / `x-forwarded-access-token`. As written, every app copying this snippet shows `userId: null`. Suggest `req.header("x-forwarded-user")`.

skills/data-app-design/references/genie-ai-trust.md
```diff
@@ -0,0 +1,67 @@
...
+
+> **GATE:** apply ONLY if the app exposes a Genie / chat / natural-language query surface. If it's a
+> dashboard / report / KPI app with no NL/chat input, do not load or apply any pattern here.
+
+An AI/Genie answer is only trustworthy if the user can **see how it was produced** and **who it ran as**.
+For ANY Genie / chat / natural-language data surface, ship ALL FIVE below — "use `GenieChat` and show
+a spinner" is not enough. Snippets use the real `@databricks/appkit` + `@databricks/appkit-ui` APIs.
+
+## 1. Authenticated identity (who is asking — on-behalf-of)
+Databricks injects identity headers; expose them and show the caller. Queries run as that user (OBO).
```
simonfaltum: **[important]** My concern here: the skill tells the UI to state that queries run on-behalf-of the signed-in user (and #5 says access is governed by the user's own permissions), but nothing in the five patterns wires the config that makes that true. Per the parent genie.md, user-scoped Genie needs `user_api_scopes: [dashboards.genie]`; a default `genie()` app can run as the app's service principal. In that setup this section mandates rendering a false security claim, which is worse than showing nothing. I think the checklist should include the `user_api_scopes` wiring, or the copy should be conditional (disclose SP execution when OBO isn't configured).

skills/data-app-design/references/genie-ai-trust.md
```diff
@@ -0,0 +1,67 @@
...
+    userId: req.header("x-databricks-user-id") ?? null,
+  });
+});
+```
+```tsx
+const [me, setMe] = useState<{ email?: string } | null>(null);
+useEffect(() => { fetch("/api/whoami").then(r => r.json()).then(setMe).catch(() => {}); }, []);
+<Badge variant="secondary">{me?.email ?? "Signed in"}</Badge>
+```
+Configure the space explicitly server-side: `genie({ spaces: { default: process.env.DATABRICKS_GENIE_SPACE_ID } })`.
```
simonfaltum: **[minor]** This conflicts with the parent genie.md, which says to call `genie()` with no arguments and let it read `DATABRICKS_GENIE_SPACE_ID` from the env. An agent loads both files at once and gets contradictory instructions. I'd defer to the parent for the default case.

skills/data-app-design/references/genie-ai-trust.md
```diff
@@ -0,0 +1,67 @@
...
+  <Card>
+    <CardHeader><CardTitle>Generated SQL</CardTitle>
+      <CardDescription>{lastSql.title ?? "How this answer was computed"}</CardDescription></CardHeader>
+    <CardContent><pre className="overflow-auto text-xs">{lastSql.query}</pre></CardContent>
+  </Card>
+)}
+```
+
+## 3. Streaming / status — not a frozen spinner
+`useGenieChat().status ∈ "idle" | "loading-history" | "streaming" | "error"`. Reflect it:
```
simonfaltum: **[nit]** Pinning the status enum (and the attachment shapes above) is the same drift class as the appkit version that got removed from the cheatsheet. genie.md defers `useGenieChat` API details to `npx @databricks/appkit docs`; I'd do the same here and keep just the pattern.

skills/data-app-design/references/appkit-cheatsheet.md
```diff
@@ -0,0 +1,52 @@
+# AppKit binding — design → primitive map
+
+This skill decides **what** to build; AppKit owns the **exact** API. For component names, props,
+hook signatures, plugin config, and the live design-token set, read the AppKit docs shipped inside
+the app — do **not** rely on a pinned list here, it drifts with the package version:
+
+- `./node_modules/@databricks/appkit-ui/CLAUDE.md` — UI: charts, tables, Genie, shadcn primitives, hooks
```
simonfaltum: **[minor]** The parent skill mandates `npx @databricks/appkit docs` as the source of truth ("ALWAYS start here", "DO NOT guess doc paths"), while this points agents at `node_modules` CLAUDE.md paths. Two different mechanisms for the same docs; I'd align with the parent's `npx` workflow so agents only learn one.

skills/data-app-design/SKILL.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Workflow
+1. **Frame** — audience, the decision/question, refresh cadence, device, primary task. One sentence.
+2. **Genre** — pick the closest from `dashboard-patterns.md` (static / analytic / magazine / infographic / repository / embedded mini). State it.
+3. **Compose** — choose content + composition patterns (data abstraction, meta-info, layout, interaction, color). Make the tradeoff explicit: what's summarized, hidden, paginated, or made interactive — and why.
+4. **Apply notation** — run the relevant `ibcs-notation.md` rules: message-in-title, scenario marks (actual/PY/plan/forecast), honest scales, semantic color. On any chart-vocabulary conflict, **IBCS wins** (see the conflict note in that file).
+5. **Bind to components** — map every element to a real primitive from `appkit-cheatsheet.md`. **Reuse `KpiCard`/`MetricTrendCard`/`HistoricalTrendCard`/`DistributionCard` before inventing** — they already encode the notation. Use `colorPalette` + semantic tokens, never hardcoded hex. Bind data with `useAnalyticsQuery`/`queryKey` + `sql.*` params.
```
simonfaltum: **[minor]** This reuse instruction is unguarded, while the cheatsheet correctly gates it on "When the app already ships these". A fresh `databricks apps init` scaffold doesn't ship `KpiCard` / `MetricTrendCard` / etc., so an agent following step 5 can import components that don't exist. Suggest carrying the same guard here: if present, reuse; otherwise create them following the notation rules.

skills/data-app-design/SKILL.md
```diff
@@ -0,0 +1,72 @@
+---
+name: data-app-design
+description: Design the UX of Databricks analytics/BI/AI data apps — dashboards, KPI/overview pages, reports, metric pages, charts, tables, AND conversational/chat/Genie natural-language data assistants — and map it to concrete AppKit components. Use when BUILDING or reviewing any UI that displays data or answers data questions: choosing genre, layout, charts, KPIs, semantic color, required states (loading/empty/error), IBCS notation, and AI-result trust (showing generated SQL/sources for Genie/chat). Complements databricks-apps (which scaffolds/builds/deploys the app); use it alongside databricks-apps whenever the app has a dashboard, chart, table, KPI, report, or Genie/chat/AI surface. NOT for non-data frontend (forms, settings, auth, marketing pages).
+metadata:
+  version: 0.1.0
+  parent: databricks-apps
```
simonfaltum: **[nit]** Convention nit: every other skill (and the README subskill example) declares `parent:` as a top-level frontmatter key, not under `metadata:`. As far as I can tell nothing consumes the field today, but worth matching the convention before tooling starts reading it.

simonfaltum (commented):
Did a deeper pass on the new skill (with a codex second opinion on top). The content and the eval setup are genuinely strong. Two important items in `genie-ai-trust.md` (an identity header that doesn't appear to exist, and an OBO claim the skill never wires up), the rest is consistency. Inline comments below.

Two things I can't attach inline because they're outside the diff:

- README's "Available Skills" section enumerates the stable skills and should include `data-app-design`.
- A net-new v0.1.0 skill lands directly in stable, which ships in every default install at the next release; the repo has `experimental/` for the maturation path. The eval rigor here is well above the bar so I'm fine either way, I just think it should be a deliberate call (lennartkats-db's, as code owner).

(Written by Isaac.)

manifest.json
```diff
@@ -1,6 +1,21 @@
 {
   "//": "GENERATED FILE: DO NOT EDIT. Run `python3 scripts/skills.py generate` to regenerate.",
   "skills": {
+    "data-app-design": {
```
simonfaltum: **[important]** This skill isn't registered in `SKILL_METADATA` (`scripts/skills.py:27`) and `.claude-plugin/plugin.json` has no matching keyword; the other 9 stable skills have both. The validate check skips skills with no metadata entry (`validate_plugin_manifests` hits `continue` with a comment saying `generate_manifest` will flag it, which it doesn't), so CI stays green while the Claude marketplace keywords silently go stale, the exact case that function's docstring warns about. CONTRIBUTING's new-skill workflow also lists the `SKILL_METADATA` step for stable skills. Fix: add a `"data-app-design": {"plugin_keyword": ...}` entry and the keyword to `plugin.json` `keywords`. (The validate gap itself deserves a tiny follow-up so the next skill PR can't miss this silently.)

skills/data-app-design/SKILL.md
```diff
@@ -0,0 +1,72 @@
+---
+name: data-app-design
```
simonfaltum: **[important]** Naming: all 9 stable skills are `databricks-` prefixed, CLAUDE.md's hierarchy names niche children `databricks-apps-*`, and CONTRIBUTING's scaffold example is `databricks-foo`. This would be the first unprefixed skill in the default-install set (the only other one, `spark-python-data-source`, sits in experimental and is named after the Spark feature). My concern is partly consistency (it renders as "Data App Design" in the middle of the "Databricks ..." list in marketplaces), partly namespace: skills land in flat namespaces across four agent ecosystems, and a generic `data-app-design` can collide with someone else's skill of the same name. Renaming after a release is painful (prompts and configs reference `$data-app-design`), so I think we should decide now. `databricks-app-design` would match the convention. The eval's routing wins came from the `databricks-apps` pointer and the description triggers rather than the name, so I wouldn't expect a rename to regress invocation, but the 5-prompt eval is cheap to re-run to confirm.

skills/data-app-design/SKILL.md
```diff
@@ -0,0 +1,72 @@
...
+description: Design the UX of Databricks analytics/BI/AI data apps — dashboards, KPI/overview pages, reports, metric pages, charts, tables, AND conversational/chat/Genie natural-language data assistants — and map it to concrete AppKit components. Use when BUILDING or reviewing any UI that displays data or answers data questions: choosing genre, layout, charts, KPIs, semantic color, required states (loading/empty/error), IBCS notation, and AI-result trust (showing generated SQL/sources for Genie/chat). Complements databricks-apps (which scaffolds/builds/deploys the app); use it alongside databricks-apps whenever the app has a dashboard, chart, table, KPI, report, or Genie/chat/AI surface. NOT for non-data frontend (forms, settings, auth, marketing pages).
```
simonfaltum: **[minor]** The synthesized marketplace blurb for this skill truncates mid-word: `synthesize_short_description` cuts the description at ". Use when" and caps at 197 chars, and the lead-in sentence here is ~235 chars, so the manifest entry ends in "and m...". It's the only manifest entry that truncates. The breadth is eval-backed so I wouldn't cut triggers; just restructure so the first sentence stands alone under ~190 chars and the detail lives after ". Use when". For reference, the full description (~750 chars) is also about 3x `databricks-apps`', and skill descriptions ride along in every session's skill list.

skills/data-app-design/references/genie-ai-trust.md
```diff
@@ -0,0 +1,67 @@
...
+app.get("/api/whoami", (req, res) => {
+  res.json({
+    email: req.header("x-forwarded-email") ?? req.header("x-forwarded-user") ?? null,
+    userId: req.header("x-databricks-user-id") ?? null,
+  });
+});
+```
+```tsx
+const [me, setMe] = useState<{ email?: string } | null>(null);
+useEffect(() => { fetch("/api/whoami").then(r => r.json()).then(setMe).catch(() => {}); }, []);
```
simonfaltum: **[nit]** `fetch(...).then(r => r.json())` with no `r.ok` check. The apps skill's own fetch examples all check `res.ok` / `response.ok` (`references/appkit/files.md`), and agents copy this shape into fetches less cosmetic than a badge. `.then(r => (r.ok ? r.json() : null))` keeps it one line.

simonfaltum (commented):
Second pass, this time verifying the load-bearing API claims against the published `@databricks/appkit-ui` 0.41.2 (the exact docs surface the cheat-sheet points at) plus the repo tooling. I checked the existing review threads first, so nothing below re-flags them. Four new items inline: two important (a marketplace registration gap, and a naming decision I think we should make before this ships), two smaller.

Verified and clean:

- `Empty` and `Skeleton` are real appkit-ui components (package `CLAUDE.md` component index); `lucide-react` is a real dependency; `shared/appkit-types/analytics.d.ts` is exactly the `appkit generate-types` default output path.
- `--success` / `--warning` (plus `-foreground` variants) exist in appkit-ui's `dist/styles.css` with light and dark values, so the semantic-color section maps to real tokens.
- `useGenieChat({ alias })` matches `UseGenieChatOptions`; `error` is typed `string | null`, so the `<Alert>{error ?? ...}</Alert>` snippet renders a string, not an object; `attachments[].query.{title,query}` match the published types. The Generated SQL snippet is shape-correct as written.
- `useAnalyticsQuery`, `sql.*` params, `config/queries/*.sql`, the `queryKey`-or-`data` chart modes, and the three `colorPalette` values all match the apps skill references.
- The published package ships a root `CLAUDE.md`, so the `node_modules` doc pointers resolve as written (the npx-vs-node_modules alignment question from the earlier review still stands).
- `python3 scripts/skills.py validate` passes at head; all relative links between the five new files resolve; the IBCS URL returns 200; no real IDs, credentials, broad grants, or destructive ops anywhere in the examples.

Net: before merge I'd fix the `SKILL_METADATA`/keyword gap, settle the naming question, and close out the two important items from the earlier review (identity header, OBO wiring). The rest is polish. The eval rigor here remains the high-water mark for the repo.

(Written by Isaac.)

skills/data-app-design/references/genie-ai-trust.md
simonfaltum: Hard evidence for this one: in the published `@databricks/appkit-ui` 0.41.2, `GenieChatStatus` is `"idle" | "loading-history" | "loading-older" | "streaming" | "error"` (`dist/react/genie/types.d.ts`), so the enum pinned here is already missing `"loading-older"`. The pin drifts in practice, not just in theory. (Written by Isaac.)

skills/data-app-design/references/genie-ai-trust.md
pkosiec: Fixed in 340960a — dropped `x-databricks-user-id` (confirmed against the Databricks Apps `http-headers` docs that no such header exists) and switched the snippet to the real headers: `x-forwarded-email` for the email and `x-forwarded-user` for the user identifier.

(Written by Isaac.)

skills/data-app-design/references/genie-ai-trust.md
pkosiec: Fixed in 340960a. The skill no longer asserts OBO unconditionally: section 1 + the checklist now require `user_api_scopes: [dashboards.genie]` (pointing at the parent `genie.md`) for on-behalf-of execution, and the disclosure is conditional — claim OBO only when that's wired, otherwise disclose the query runs as the app's service principal. Applied the same conditional framing in `SKILL.md` so the two files don't contradict.

(Written by Isaac.)

skills/data-app-design/references/genie-ai-trust.md
pkosiec: Fixed in 340960a — defer to the parent default: register with no-arg `genie()` (reads `DATABRICKS_GENIE_SPACE_ID` from the env). Dropped the pinned `genie({ spaces: {...} })` shape and point multi-space config at `npx @databricks/appkit docs`.

(Written by Isaac.)

skills/data-app-design/references/genie-ai-trust.md
pkosiec: Agreed — and thanks for the 0.41.2 evidence (`loading-older`). Fixed in 340960a: dropped the pinned `GenieChatStatus` union and the attachment-shape enumeration, deferring both to `npx @databricks/appkit docs "useGenieChat"` and keeping just the pattern.

(Written by Isaac.)

skills/data-app-design/references/appkit-cheatsheet.md
pkosiec: Fixed in 340960a — the cheatsheet now points at `npx @databricks/appkit docs` (the parent's mandated source of truth, "ALWAYS start here") instead of the `node_modules` CLAUDE.md paths, so agents learn one mechanism.

(Written by Isaac.)

skills/data-app-design/SKILL.md
pkosiec: Fixed in 340960a — carried the cheatsheet's guard into step 5: reuse `KpiCard`/`MetricTrendCard`/… only if the app already ships them; a fresh scaffold won't, so create them following the notation rules.

(Written by Isaac.)

skills/data-app-design/SKILL.md
pkosiec: Fixed in 340960a — moved `parent: databricks-apps` to a top-level frontmatter key (kept `version` under `metadata:`), matching the convention.

(Written by Isaac.)

manifest.json
pkosiec: Fixed in 340960a. Added `"databricks-app-design": {"plugin_keyword": "app-design"}` to `SKILL_METADATA` and the `app-design` keyword to `plugin.json`. Also closed the validate gap you flagged: `validate_plugin_manifests` no longer silently `continue`s past a stable skill with no metadata entry — it appends a hard error now (the old comment claiming `generate_manifest` would catch it was false). Verified it fires by dropping the entry locally.

(Written by Isaac.)

skills/data-app-design/SKILL.md
pkosiec: Done in 340960a — renamed to `databricks-app-design` to match the stable-set convention and avoid a flat-namespace collision. Touched the dir, frontmatter `name`, the `databricks-apps` cross-ref and the Codex `openai.yaml` prompt token; regenerated the manifest; `grep` confirms zero stale `data-app-design` refs.

(Written by Isaac.)

skills/data-app-design/SKILL.md
pkosiec: Fixed in 340960a — rewrote the description so the lead sentence stands alone at 158 chars before ". Use when"; the synthesized manifest blurb now ends cleanly ("…concrete AppKit components.") instead of mid-word.

(Written by Isaac.)

skills/data-app-design/references/genie-ai-trust.md
pkosiec: Fixed in 340960a — `fetch("/api/whoami").then(r => (r.ok ? r.json() : null))`.

(Written by Isaac.)

skills/data-app-design/references/genie-ai-trust.md
pkosiec: Small correction to my note above: the working pointer is `npx @databricks/appkit docs ./docs/plugins/genie.md` (the `useGenieChat` hook is documented in the Genie plugin doc) — a bareword `docs "useGenieChat"` doesn't resolve against appkit 0.41.2. Corrected across the skill in c148b99.

(Written by Isaac.)

pkosiec:
@simonfaltum @MarioCadenas — review addressed; all inline threads resolved. Final commit after an amend is **c148b99** (history was force-updated, so the earlier `340960a` references point at the pre-amend commit).

**Important**
- Real `x-forwarded-email` / `x-forwarded-user` headers (dropped the non-existent `x-databricks-user-id`).
- OBO is conditional on `user_api_scopes: [dashboards.genie]` (else discloses service-principal execution) — in both `genie-ai-trust.md` and `SKILL.md`.
- Skill registered: `SKILL_METADATA` + `app-design` keyword; `validate_plugin_manifests` now errors on an unregistered stable skill instead of silently passing.
- Renamed `data-app-design` → **`databricks-app-design`**.

**Smaller:** `genie()` defers to the parent no-arg default; no pinned status enum / attachment shapes; guarded component reuse; `parent:` moved top-level; `r.ok` check; description no longer truncates mid-word.

**Doc pointers verified against appkit 0.41.2** (`appkit docs` installed locally): every `npx @databricks/appkit docs` target resolves — section names (`appkit-ui API reference`) and doc paths (`./docs/plugins/genie.md`, `./docs/plugins/analytics.md`). Bareword queries (`"genie"`, `"useGenieChat"`, `"analytics"`) don't resolve, so those were corrected. Also aligned the chart section to the parent `frontend.md` (ECharts components, `parameters` required in query mode) and re-confirmed the named UI primitives + semantic tokens exist.

**Out-of-diff**
- Added `databricks-app-design` to the README "Available Skills" list.
- Routing (your standalone question): added a when-NOT carve-out excluding managed AI/BI (Lakeview) dashboard authoring. The broader "App vs AI/BI" product-decision gate in `databricks-apps` is the right long-term fix but bigger than this PR — worth its own follow-up issue (happy to open one).

**Needs a call:** keeping this as a net-new **stable** skill (vs `experimental/`) is deliberate given the eval rigor, but it ships in every default install — @lennartkats-db, requesting your sign-off as code owner before merge.

(Written by Isaac.)

pkosiec (approved):
Pushed fixes addressing @simonfaltum's comments + also fixed usage of unexported Dev Playground components. Now it looks good to me 👍

simonfaltum (approved):

PR #128, opened 2026-06-08
feat(plugin): add prompt-routing hooks and doctor/setup commands

simonfaltum:
## Why

The plugin shipped skills only. The skills carry the Databricks judgment, but nothing made sure a Databricks-related prompt actually reached them, and there were no entry points for common setup or health-check tasks. This adds a thin hook and command layer so Databricks work routes into the skills automatically, staying CLI-first (no MCP).

## What the hooks do (plain language)

The hooks make sure Databricks work goes through the skills instead of Claude improvising:

- When your prompt mentions something Databricks-related, the **router** hook adds a short note to that turn telling Claude to load the relevant skill(s) first, and it stays silent on non-Databricks prompts. The full note fires once per session; later Databricks prompts get a one-line reminder, so long sessions don't pay the full block every turn.
- The **context** hook runs at session start (new sessions, `/clear`, and after compaction, but not on resume where the banner is already in context) and injects the routing reminder plus a one-line CLI/profile status.
- The **auth helper** hook watches `databricks` command results; when one fails with an auth-shaped error (expired/invalid token, missing credentials, OAuth refresh failure), it adds one line suggesting `/databricks:doctor` or `databricks auth login` before any retry.

All three only *inject context* that Claude then acts on (technically `additionalContext`, a just-in-time nudge, not a forced skill load). They never gate, block, or change your commands, there are no permission prompts or cost warnings, and they fail open (any error means no output, never a blocked prompt or tool call).

## Changes

Before: skills only. Now: skills + three hooks + two commands.

Hooks (`hooks/`, auto-loaded via `hooks/hooks.json`, all fail open, stdlib only):
- `databricks-router.py` (UserPromptSubmit): a sub-50ms keyword regex over the prompt. On a Databricks match it injects context steering Claude to load `databricks-core` plus the matching product skill; the full instruction is injected once per session (marker file keyed on `session_id`), then a one-line reminder. Unrelated prompts are left untouched. Code-hosting URLs (`github.com/databricks/...`) do not count as Databricks intent; URLs whose hostname contains `databricks` (workspace, docs) do.
- `databricks-context.py` (SessionStart, matcher `startup|clear|compact`): injects the routing rule plus CLI version, configured profile names, and any `[__settings__].default_profile` (read locally, no network call, no token values).
- `databricks-auth-helper.py` (PostToolUse, matcher `Bash`): phrase-based auth-failure detection on `databricks` command output (`cannot configure default credentials`, `invalid_grant`, `401 unauthorized`, invalid/expired token); injects a one-line fix suggestion. Bare status codes in ordinary data never trip it, and only commands that actually invoke the `databricks` executable count (a repo path or URL containing `databricks`, like `gh pr view --repo databricks/cli`, does not).

Commands (`commands/`):
- `/databricks:doctor`: read-only health check (CLI, profiles, auth method via `auth describe`, auth validity via a single `current-user me` call). Compute and recent-run-failure checks only run with the `full` argument.
- `/databricks:setup`: auth onboarding (OAuth / PAT / service principal), covering workspace URLs, the no-host login.databricks.com flow, and account console URLs with `--account-id`. Interactive auth commands (PAT paste, browserless OAuth) are handed to the user's own terminal instead of being run without a TTY.

Product workflows stay in the skills, so no command shadows a skill of the same name. `plugin.json` declares `commands`; `hooks/hooks.json` is auto-loaded by Claude Code and intentionally not declared (declaring the standard path double-loads it and fails the plugin).

Tooling: `scripts/skills.py validate` now validates the components (hooks.json is valid JSON referencing scripts that exist, every command has a description without unquoted-colon YAML pitfalls, and plugin.json does not re-declare the auto-loaded hooks file, string or list form). CI runs the hook test suite from `tests/` with a single `unittest discover` command. README, CONTRIBUTING, and CLAUDE.md document the new surface.

## Test plan

- [x] `python3 -m unittest discover -s tests -p '*_test.py'` (36 tests: strong / ambiguous / suppressed routing including declarative pipelines vs Jenkins, URL handling, once-per-session memo, payload shapes, config parsing, default_profile, sanitization, auth-failure phrases, executable-token command detection)
- [x] `python3 scripts/skills.py validate`
- [x] End-to-end via the hook stdin/stdout contract (router full-then-reminder across invocations, auth hint vs silent, context primer output, fail-open on malformed input)
- [x] Installed locally and verified in-session: `/reload-plugins` loads the hooks, the router injects context live, both commands register
- [x] Command frontmatter parses under strict YAML (PyYAML `safe_load`)

## Distribution & releases

How this reaches users, and how it relates to releases:

- **Claude Code marketplace** (`databricks/databricks-agent-skills`, `source: "./"`): tracks `main`, but clients only pick up a change once the **`version`** field in `.claude-plugin/plugin.json` is bumped, which happens only via the release workflow (`bump_version.py` commits the bump to `main` and tags `vX.Y.Z`). **This PR intentionally does not bump `version`** (the release workflow owns that), so these hooks/commands land on `main` at merge but reach marketplace users only after the next release is cut, not on merge.
- **`databricks aitools install`** (CLI path): resolves via `cli-compat.json` in `databricks/cli` and currently packages **skills only**. Shipping `hooks/` + `commands/` through this path is separate CLI-side follow-up work.

Ship path: merge this, then run the release workflow for a new `vX.Y.Z` (bumps `version` + tags), and marketplace clients get it on `marketplace update`.

### DIFF
--- a/.github/workflows/validate-manifest.yml
+++ b/.github/workflows/validate-manifest.yml
@@ -33,3 +33,6 @@ jobs:
 
       - name: Test databricks-router hook
         run: python3 hooks/databricks_router_test.py
+
+      - name: Test databricks-context hook
+        run: python3 hooks/databricks_context_test.py
--- a/.github/workflows/validate-manifest.yml
+++ b/.github/workflows/validate-manifest.yml
@@ -11,6 +11,7 @@ on:
       - '.claude-plugin/**'
       - 'hooks/**'
       - 'commands/**'
+      - 'tests/**'
   push:
     branches:
       - main
@@ -32,7 +33,4 @@ jobs:
         run: python3 scripts/skills.py validate
 
       - name: Test plugin hooks
-        run: |
-          python3 hooks/databricks_router_test.py
-          python3 hooks/databricks_context_test.py
-          python3 hooks/databricks_auth_helper_test.py
+        run: python3 -m unittest discover -s tests -p '*_test.py' -v
--- a/.github/workflows/validate-manifest.yml
+++ b/.github/workflows/validate-manifest.yml
@@ -31,8 +31,8 @@ jobs:
       - name: Validate manifest is up to date
         run: python3 scripts/skills.py validate
 
-      - name: Test databricks-router hook
-        run: python3 hooks/databricks_router_test.py
-
-      - name: Test databricks-context hook
-        run: python3 hooks/databricks_context_test.py
+      - name: Test plugin hooks
+        run: |
+          python3 hooks/databricks_router_test.py
+          python3 hooks/databricks_context_test.py
+          python3 hooks/databricks_auth_helper_test.py
--- a/.github/workflows/validate-manifest.yml
+++ b/.github/workflows/validate-manifest.yml
@@ -9,6 +9,8 @@ on:
       - 'scripts/skills.py'
       - 'manifest.json'
       - '.claude-plugin/**'
+      - 'hooks/**'
+      - 'commands/**'
   push:
     branches:
       - main
@@ -28,3 +30,6 @@ jobs:
 
       - name: Validate manifest is up to date
         run: python3 scripts/skills.py validate
+
+      - name: Test databricks-router hook
+        run: python3 hooks/databricks_router_test.py
--- a/commands/doctor.md
+++ b/commands/doctor.md
@@ -18,8 +18,10 @@ report the rest as unknown.
 2. **Profiles**: `databricks auth profiles`. List configured profiles and
    validity. If `$1` is given, use that profile for the rest. Otherwise, if more
    than one profile exists, ask the user which to use (**never auto-select**).
-3. **Identity**: `databricks current-user me --profile <profile>`. Confirm who
-   you're authenticated as.
+3. **Identity & auth method**: `databricks auth describe --profile <profile>`
+   shows the effective host, user, and credential source (never pass
+   `--sensitive`). Fall back to `databricks current-user me --profile <profile>`
+   if it's unavailable.
 4. **Workspace reach**: `databricks catalogs list --profile <profile>`.
    Confirms API reachability and Unity Catalog access.
 5. **Compute**: `databricks warehouses list` and `databricks clusters list` for
--- a/commands/doctor.md
+++ b/commands/doctor.md
@@ -13,7 +13,8 @@ check `databricks <group> --help` first rather than guessing.
 Run these in order. Don't stop on the first failure — collect what you can and
 report the rest as unknown.
 
-1. **CLI** — `databricks --version`. Flag if missing or below v0.292.0.
+1. **CLI** — `databricks --version`. Flag only if it's missing; don't gate on a
+   specific version (the CLI surfaces its own update notice).
 2. **Profiles** — `databricks auth profiles`. List configured profiles and
    validity. If `$1` is given, use that profile for the rest. Otherwise, if more
    than one profile exists, ask the user which to use — **never auto-select**.
--- a/commands/doctor.md
+++ b/commands/doctor.md
@@ -1,5 +1,5 @@
 ---
-description: Read-only Databricks health check — CLI version, auth, workspace reachability, running compute, recent job failures.
+description: Read-only Databricks health check: CLI version, auth, workspace reachability, running compute, recent job failures.
 argument-hint: "[profile]"
 allowed-tools: Bash(databricks:*), Read
 ---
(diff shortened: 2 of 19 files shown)


commands/doctor.md
```diff
@@ -0,0 +1,36 @@
...
+   shows the effective host, user, and credential source (never pass
+   `--sensitive`). Fall back to `databricks current-user me --profile <profile>`
+   if it's unavailable.
+4. **Workspace reach**: `databricks catalogs list --profile <profile>`.
+   Confirms API reachability and Unity Catalog access.
+5. **Compute**: `databricks warehouses list` and `databricks clusters list` for
+   the profile. Note what's running.
+6. **Recent job failures**: list recent job runs (e.g.
+   `databricks jobs list-runs --limit 20 --profile <profile>`) and surface any
+   recent failures.
```
parthban-db: Why do we have these checks? What value does it provide? Because if we want to see whether the profile is valid, we can simply test a single API.

commands/setup.md
```diff
@@ -0,0 +1,40 @@
+---
+description: "Set up Databricks CLI auth: install check, then an OAuth / PAT / service-principal profile, then verify."
+argument-hint: "[workspace-url]"
```
parthban-db: What about spog URLs/account ID?

hooks/databricks-auth-helper.py
```diff
@@ -0,0 +1,127 @@
...
+    except Exception:
+        result = None
+
+    if result:
+        print(json.dumps({
+            "hookSpecificOutput": {
+                "hookEventName": "PostToolUse",
+                "additionalContext": result,
+            }
+        }))
```
parthban-db: Not sure, but can it error out? To be safe, should we cover the entire main block with a try cache?

.github/workflows/validate-manifest.yml
```diff
@@ -28,3 +30,9 @@ jobs:
 
       - name: Validate manifest is up to date
         run: python3 scripts/skills.py validate
+
+      - name: Test plugin hooks
+        run: |
+          python3 hooks/databricks_router_test.py
+          python3 hooks/databricks_context_test.py
+          python3 hooks/databricks_auth_helper_test.py
```
parthban-db: Can we have the test in a separate test directory? and run the complete suite with a single command?

parthban-db (approved):

hooks/README.md
```diff
@@ -0,0 +1,76 @@
...
+the full routing block on every turn.
+
+There's no second agent to delegate to. Claude itself drives the `databricks`
+CLI through the skills, so "routing" just means "make sure the Databricks skills
+are loaded." There is **no permission gating and no cost warning** here.
+
+Precision is tuned to avoid over-routing:
+
+- **STRONG** terms (`databricks`, `unity catalog`, `lakeflow`, `dbfs`,
+  `databricks.yml`, `delta live tables`, `genie`, ...) always route, even
```
dustinvannoy-db: Should change `delta live tables` to `declarative pipelines`

dustinvannoy-db (approved):
One comment to consider and messaged about a design decision. Nothing blocking this PR from my perspective.

commands/doctor.md
simonfaltum: Fair point. Reworked in c97e670: auth validity is now exactly one API call (`databricks current-user me`), and the default run stops there (CLI, profiles, `auth describe`, that one probe). The catalogs call is gone. Compute and recent-job-failure checks now only run when the user passes `full` or explicitly asks, for the case where someone wants an environment snapshot rather than an auth check.

commands/setup.md
simonfaltum: Good catch, covered in c97e670. With no URL, setup now uses the CLI's no-host flow (`databricks auth login` opens login.databricks.com to sign in and pick a workspace), and browser-copied URLs carrying `?w=` / `account_id` query params are accepted (quoted so the shell doesn't eat the `?`). Account console URLs (`accounts.cloud.databricks.com` etc.) get `--account-id <uuid>`, and the verify step notes that `current-user me` is workspace-only, so account profiles verify via `auth describe`.

hooks/databricks-auth-helper.py
simonfaltum: It could, in theory. Parsing and the check itself were guarded, but the final json.dumps/print sat outside any try, so something like a closed stdout would have surfaced as a hook error. Fixed in c97e670: all three hooks now build their output inside one outer try covering the whole main block, and the final print has its own guard, so any error still ends as `{}` / no output with exit 0.

.github/workflows/validate-manifest.yml
simonfaltum: Done in c97e670: tests moved to `tests/` at the repo root, and the whole suite runs with a single command, `python3 -m unittest discover -s tests -p '*_test.py'` (36 tests). CI now uses that command.

hooks/README.md
simonfaltum: Done in c97e670. The router now routes `declarative pipelines`: `spark declarative pipelines` is a strong term, and the bare phrase sits in the ambiguous tier with `jenkins`/`Jenkinsfile` as new suppress terms, so Jenkins declarative pipelines stay out. The routing instruction maps it to databricks-pipelines under the new name, and `delta live tables` stays only as a legacy detection alias (plenty of prompts still use the old name), marked as such in the README.

PR #133, opened 2026-06-09
feat(plugin): add databricks-experimental as a skills-only plugin

simonfaltum:
> **Stacked on #128.** Base is `simonfaltum/plugin-hooks-commands`; only the last commit is new here. Merge #128 first, then retarget this to `main`.

## Why

Experimental skills are installable today only through the CLI's files overlay (`databricks aitools install --experimental`), so they never get the out-of-band updates the plugin channel provides, and graduating a skill to stable leaves CLI users with duplicate copies to clean up. The aitools plugin-first proposal moves experimental content to its own plugin in this repo's marketplace so the agent keeps it updated and graduation becomes a pure file move.

The plugin is deliberately **skills-only**, and that is a policy, not a gap: Claude Code runs hooks from every enabled plugin with no shadowing or precedence, so any hook shipped here would stack with the stable plugin's hooks (#128) on every prompt, session start, and tool call. The stable `databricks` plugin is the platform layer (prompt router, context primer, auth hints, slash commands); `databricks-experimental` is a content pack that assumes the stable plugin is installed alongside it. This also means weekly-churn experimental code never runs in the always-on hook hot path.

## Changes

Before: one plugin (`databricks`) in the marketplace; experimental skills reachable only via the CLI files overlay. Now: a second plugin, `databricks-experimental`, in the same self-hosted marketplace.

- `.claude-plugin/marketplace.json`: second entry, `source: "./experimental"`.
- `experimental/.claude-plugin/plugin.json` (new): name `databricks-experimental`, `"skills": "./"` (the skill dirs sit directly at the plugin root; there is no `experimental/skills/`), version `0.2.0` in lockstep with the stable manifest, and a description that states the content is unstable and may change or be removed.
- `scripts/skills.py`: new `check_experimental_plugin` in `validate` enforces the policy: manifest exists, parses, correct name, `"skills": "./"` present (without it the plugin silently ships zero skills), no `hooks`/`commands`/`agents`/`mcpServers` keys and no `experimental/hooks|commands|agents` dirs, version lockstep with the stable manifest, marketplace source correct. Marketplace description sync now covers both plugins.
- `scripts/bump_version.py`: the release bump now sets the experimental manifest too, so one tag means one version across all plugin manifests.
- Docs: README (marketplace install for both plugins, updated CLI vs marketplace table), `experimental/README.md` (plugin install + the skills-only rule), CONTRIBUTING ("The experimental plugin is skills-only" policy section, release step update), CLAUDE.md.

Notes:

- The plugin stays on this repo's self-hosted marketplace only. It is never submitted to Anthropic's community or official listings; the stable plugin is the candidate for those.
- Skill-name collisions between `skills/` and `experimental/` are already rejected at manifest generation, which keeps graduation atomic. Users who update the two plugins at different times can briefly see a graduated skill twice (namespaced `databricks:<name>` and `databricks-experimental:<name>`); updating both plugins resolves it, and `aitools update` on the plugin channel will do that in one go (CLI-side work, tracked in the aitools plugin-first proposal).
- An explicit `version` field pins marketplace updates to release bumps, matching the stable plugin's release-gated model (omitting it would make every commit to `main` an update, which conflicts with release gating).

## Test plan

- [x] Negative tests: validator rejects an `experimental/hooks/` dir and a version skew between the two manifests (verified manually, then reverted)
- [x] `python3 scripts/bump_version.py v0.2.0` smoke test: reports all three manifests, regenerates manifest.json as a no-op (cursor side effect reverted; the release workflow owns that bump)
- [x] `claude plugin validate` passes on both the plugin manifest and the marketplace manifest
- [x] Real install in an isolated `CLAUDE_CONFIG_DIR`: `claude plugin marketplace add <repo>` + `claude plugin install databricks-experimental@databricks-agent-skills` run headlessly with no prompts; `claude plugin details` shows all 17 experimental skills, 0 hooks, 0 commands, ~1.7k always-on tokens
- [x] Stable plugin installed alongside in the same config: inventory unchanged (skills + 2 commands + 3 hooks, no experimental leakage), both plugins coexist
- [x] All hook test suites still pass (router 16, context 11, auth-helper 8)

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -11,6 +11,12 @@
       "source": "./",
       "category": "development",
       "description": "Databricks skills for the CLI, Apps, Lakebase, Model Serving, Lakeflow Jobs, Spark Declarative Pipelines, Declarative Automation Bundles (DABs), and classic-to-serverless migration."
+    },
+    {
+      "name": "databricks-experimental",
+      "source": "./experimental",
+      "category": "development",
+      "description": "Experimental Databricks skills: best-effort, unstable content that may change or be removed without notice. Skills only by design; install the stable databricks plugin alongside it for the routing hooks and slash commands."
     }
   ]
 }
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -71,6 +71,14 @@ changing hook behavior.
 These ship via the plugin marketplace
 (whole-repo source); `databricks aitools install` currently installs skills only.
 
+`experimental/` is also its own **skills-only** plugin (`databricks-experimental`,
+manifest at `experimental/.claude-plugin/plugin.json`, marketplace source
+`./experimental`). Never add hooks or commands under `experimental/`; hooks
+stack across enabled plugins, so they live in the stable plugin only, and
+`scripts/skills.py validate` rejects violations. Both Claude plugin manifests
+share one version, bumped in lockstep by `scripts/bump_version.py` on release.
+See CONTRIBUTING.md ("The experimental plugin is skills-only").
+
 ## Security
 
 When documenting examples, obfuscate sensitive info:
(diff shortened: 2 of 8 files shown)


dustinvannoy-db (approved):
LGTM

PR #138, opened 2026-06-11
feat(cursor): ship commands and hooks in the Cursor plugin

simonfaltum:
## Why

The Cursor marketplace plugin (`databricks-skills`, live since Feb 2026) ships skills only, while the Claude Code plugin also carries the setup/doctor commands and three hooks. Cursor supports plugin commands and hooks, so we can close most of that gap in place. The plugin name stays `databricks-skills`: it is the frozen Cursor install identifier (see `.cursor-plugin/NOTES.md`), so this PR is strictly additive.

## Changes

Before: Cursor users got the 9 skills, no commands, no hooks. Now: the same plugin also ships `/databricks-setup` and `/databricks-doctor` plus two hooks (session context primer, auth-failure hint).

- `.cursor-plugin/plugin.json`: adds `"commands": "./commands-cursor/"` and `"hooks": "./hooks/cursor-hooks.json"`. The explicit hooks pointer is load-bearing: without it Cursor default-discovers `hooks/hooks.json`, which is the Claude-format wiring.
- `commands-cursor/`: Cursor copies of the two commands. Cursor has a flat `/` menu (no `databricks:` namespacing), so they are named `databricks-setup` / `databricks-doctor`, and `$1` argument references become plain prose (Cursor has no documented args templating).
- `hooks/cursor-hooks.json`: wires `sessionStart` (context primer) and `postToolUse` with matcher `Shell` (auth hinter), both invoked with `--platform cursor`.
- `hooks/databricks-context.py`, `hooks/databricks-auth-helper.py`: learn a `--platform` flag (default `claude`, unknown values fall back to `claude` so a wiring typo degrades instead of breaking). It switches the output envelope (`hookSpecificOutput.additionalContext` vs Cursor's `additional_context`) and the command names rendered in injected text. The auth hinter also accepts the shell tool names of other harnesses (`Shell`, `run_in_terminal`, `unified_exec`, ...) and parses Cursor payloads where `tool_input` / `tool_output` arrive as JSON-encoded strings.
- The prompt router is intentionally not ported: Cursor's `beforeSubmitPrompt` hook can block a prompt but cannot inject context. Routing rides on the primer's routing rule plus Cursor's native skill selection.
- `scripts/skills.py` validate: new `check_cursor_plugin` guard. Pins the frozen plugin name, requires the explicit `cursor-hooks.json` pointer, verifies referenced hook scripts exist, and checks command frontmatter.
- `README.md` + `hooks/README.md`: document the Cursor surface.

## Test plan

- [x] `python3 -m unittest discover -s tests -p '*_test.py'` (52 tests, including new platform/envelope/payload coverage)
- [x] `python3 scripts/skills.py validate`
- [ ] Manual verification in Cursor before release: plugin-hook script paths are documented as project-root/user-dir relative but plugin-relative resolution is undocumented, and command exec semantics (shell vs spawn) are undocumented. If paths resolve differently the wiring file needs a path prefix, not a script change.

Note: `cursor-agent` (CLI) plugin loading is undocumented; this targets the IDE surface first.


commands-cursor/databricks-doctor.md
```diff
@@ -0,0 +1,41 @@
+---
+name: databricks-doctor
```
parthban-db, line 2: Why did you copy the commands instead of making the symlinks? Also, if necessary, can we do something to ensure they are in sync? I would prefer if it could be a static script, but an agent workflow is also fine.

simonfaltum: They're not identical, so a symlink wouldn't hold. The Cursor copies are adapted to Cursor's plugin format:

- Frontmatter: Cursor uses `name:` and ignores the Claude-only fields (`argument-hint:`, `allowed-tools:`).
- No `$1` args templating, so `$1` becomes prose ("the profile the user named after the command").
- Flat command menu (no `databricks:` namespace), so `/databricks:setup` becomes `/databricks-setup`.

A symlink would force identical content and break all three.

On keeping them in sync: agreed a static script is the right call. The transforms above are mechanical, so `commands/` becomes the source of truth and `commands-cursor/` is generated from it. I'd rather land that as a follow-up once the other plugin surfaces (cursor, proj, experimental) are merged, so one generator covers every target instead of growing per-PR. There's already a generation path in `scripts/skills.py` (+ `.gen.json`) it can extend. Until then these two are hand-maintained; validate checks their frontmatter but not content drift.

parthban-db (approved):

PR #152, opened 2026-06-16
build(plugin): generate all plugin manifests from one plugin.meta.json

simonfaltum:
## Why

The repo ships one logical plugin to four targets (Claude Code, Codex, Copilot, Cursor) plus three marketplace catalogs. Every target hand-maintained the same plugin metadata, and the copies drifted:

- Version was duplicated across all four `plugin.json` files (`0.2.4`).
- Codex's `plugin.json` was missing the `app-design` keyword that Claude's had.
- Cursor's description had gone stale ("...Unity Catalog...") versus the canonical one.

There was no single place to change the plugin's version, identity, description, or keywords.

## Changes

**Before:** four hand-edited `plugin.json` + three `marketplace.json`, kept in sync by hand (and `bump_version.py` regex-editing each manifest).

**Now:** `plugin.meta.json` at the repo root is the single source of truth, and `scripts/skills.py` generates all four `plugin.json` + three `marketplace.json` from it. CI re-renders them in memory and fails on any byte drift.

- `plugin.meta.json` (new): version, identity, description, author/urls/license, composed keywords, per-target config (display names, the Codex `interface`, hook/command/rule wiring), and the `skills` map (skill to plugin keyword, folding in the old `SKILL_METADATA`).
- `scripts/skills.py`: per-target builders, `generate_plugins`, a byte-drift check, and a skill-coverage check (every skill must have a meta entry and vice versa). Removed `SKILL_METADATA` and `validate_plugin_manifests` (subsumed). Kept the existing per-target hook/command/routing checks as defense-in-depth.
- `scripts/bump_version.py`: writes the version into `plugin.meta.json`, then regenerates everything, so all four targets always share one version.
- `.github/workflows/release.yml`: commits the full generated set (previously it staged only two of the four `plugin.json`, so Codex/Copilot bumps were never committed) and runs `validate` before tagging.
- `.github/workflows/validate-manifest.yml`: expanded PR triggers to every generated dir + `plugin.meta.json` + `rules/`.
- `tests/skills_generator_test.py` (new): 13 tests pinning generation, drift detection, and coverage.
- Also fixes the `databricks-core` manifest blurb, which truncated mid-word to "...Load this firs...", by adding ". Load this" as a short-description boundary.

Generating fixes the two existing drifts as a side effect: Codex gains `app-design`; Cursor's description becomes canonical. Every other generated file is byte-identical to what shipped before.

Note: the generated `plugin.json` / `marketplace.json` carry no "do-not-edit" comment key, because the plugin loaders and the Claude marketplace `$schema` reject unknown keys. Their generated status is documented in `CONTRIBUTING.md` and enforced by the drift check.

## Test plan

- [x] `python3 scripts/skills.py generate` produces a diff of only the two intended drift-fixes; all other generated files byte-identical.
- [x] `python3 scripts/skills.py validate` passes; the drift guard fails (exit 1) on a tampered manifest.
- [x] `python3 scripts/bump_version.py vX.Y.Z` propagates the version to all four `plugin.json` and round-trips cleanly.
- [x] `python3 -m unittest discover -s tests -p '*_test.py'` (65 tests) passes.
- [x] Adversarial review (4 reviewers + per-finding verification): 6 findings, all low/nit, all addressed.

### DIFF
--- a/.agents/plugins/README.md
+++ b/.agents/plugins/README.md
@@ -1,0 +1,9 @@
+<!-- GENERATED FILE: do not edit by hand. -->
+
+# Generated plugin manifests
+
+The plugin manifest files in this directory (`plugin.json` and/or
+`marketplace.json`) are generated from the repo-root `plugin.meta.json` by
+`scripts/skills.py`. To change them, edit `plugin.meta.json` and run
+`python3 scripts/skills.py generate`. CI re-renders these files and fails on any
+drift, so hand-edits are reverted. See `CONTRIBUTING.md` ("Plugin metadata").
--- a/.claude-plugin/README.md
+++ b/.claude-plugin/README.md
@@ -1,0 +1,9 @@
+<!-- GENERATED FILE: do not edit by hand. -->
+
+# Generated plugin manifests
+
+The plugin manifest files in this directory (`plugin.json` and/or
+`marketplace.json`) are generated from the repo-root `plugin.meta.json` by
+`scripts/skills.py`. To change them, edit `plugin.meta.json` and run
+`python3 scripts/skills.py generate`. CI re-renders these files and fails on any
+drift, so hand-edits are reverted. See `CONTRIBUTING.md` ("Plugin metadata").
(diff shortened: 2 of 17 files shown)


pkosiec (approved):
LGTM

PR #153, opened 2026-06-16
feat(routing): generate routing tables from plugin.meta.json (P2)

simonfaltum:
> Stacked on #152 (base branch `simonfaltum/meta-plugin-generator`). Review/merge #152 first; this PR's diff is just the two routing commits.

## Why

The product-skill routing table was duplicated in two hand-maintained places: the prompt router's `ROUTING_INSTRUCTION` (`hooks/databricks-router.py`, used on Claude/Codex) and the Cursor rule `rules/databricks-routing.mdc`. #151 added a check that the two *agree*, but nothing made them come from a single source, and the router's keyword lists (STRONG/AMBIGUOUS/SUPPRESS) lived only in the script.

## Changes

This is P2 of the meta-plugin work, in two commits so the risky part is reviewable in isolation.

**Commit 1 (`feat(routing)`, additive):** add a `routing` block to `plugin.meta.json` (the strong/ambiguous/suppress regex lists, the product-skill table, and the preamble/closing prose). `scripts/skills.py` renders it into:
- `hooks/_routing_data.json` (new) which the router loads, and
- `rules/databricks-routing.mdc` (regenerated **byte-identical** to the hand-authored file),

so both routing tables come from one `table` and cannot drift. Adds a drift check, a structural coverage check (every stable product skill needs a `routing.table` row, derived from its frontmatter `parent`), a `rules/README.md` "generated" marker, and cross-check tests asserting the rendered data equals the router's current values. The router is untouched in this commit.

**Commit 2 (`refactor(routing)`, the cutover):** `hooks/databricks-router.py` now loads its config from `_routing_data.json` instead of hardcoding it. Guarded and fail-open: if the file is missing or unreadable, a minimal inline fallback still routes obvious `databricks` prompts. `ROUTING_INSTRUCTION` stays a module attribute (the CI routing check reads it) and every public function is unchanged.

Routing behavior is identical: the cross-check tests prove the generated data equals the prior inline lists/instruction, so the cutover swaps the *source* of the config, not its values.

## Test plan

- [x] `scripts/skills.py generate` regenerates `rules/databricks-routing.mdc` byte-identically (no diff) and writes `_routing_data.json`.
- [x] Cross-check tests: rendered `strong`/`ambiguous`/`suppress`/`instruction`/`reminder` equal the router's values; `.mdc` equals disk.
- [x] Coverage check: every stable product skill (parent none or `databricks-core`) has a routing row; `app-design` (parent `databricks-apps`) and `core` exempt.
- [x] Router suite (keyword precision + session memo) + new fallback-path tests (missing/corrupt/incomplete data file) pass: 81 tests total.
- [x] `scripts/skills.py validate` green (includes the new routing drift + coverage checks and #151's table-sync check).
- [x] End-to-end smoke: piping a prompt to the hook emits the full instruction loaded from the data file.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -60,6 +60,12 @@ config, edit `plugin.meta.json` and run `python3 scripts/skills.py generate`; CI
 and propagates to all four targets. Adding a stable skill requires an entry in
 the `skills` map (with a `keyword`).
 
+The `routing` block in `plugin.meta.json` is generated too: it renders the
+prompt router's data (`hooks/_routing_data.json`, loaded by
+`hooks/databricks-router.py`) and the Cursor rule
+(`rules/databricks-routing.mdc`) from one table. Do not hand-edit those; edit
+`plugin.meta.json` and regenerate. CI fails if a product skill has no routing row.
+
 ## Plugin components (hooks + commands)
 
 Beyond skills, the Claude Code plugin ships two component dirs at the repo root.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,6 +80,13 @@ keywords_tail`, in the insertion order of the `skills` map. The plugin `name` is
 `databricks` for every target and is load-bearing (it keys Cursor/Claude
 installs); the generator never emits a different value.
 
+The `routing` block in `plugin.meta.json` is also generated output: the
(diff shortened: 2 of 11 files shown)


pkosiec (approved):

dustinvannoy-db (approved):
LGTM

PR #155, opened 2026-06-16
refactor: split the generator into a package, add a metaplugin/ source folder, close two guard gaps

simonfaltum:
## Why

Follow-up to reviewer feedback on the meta-plugin generator (#152/#153/#154) that we had not addressed:

- `scripts/skills.py` had grown to ~1750 lines in one flat module.
- The editable source was not obvious; it should read as "this is where we edit things."
- Two non-blocking guard gaps: the hook-wiring path was declared twice per target with no enforced consistency, and an orphaned `hooks/*.py` (present but never wired) was not flagged.

## Changes

Three reviewable commits:

1. **Split `scripts/skills.py` into a `scripts/skillsgen/` package** (common, discovery, manifest, plugins, routing, hooks, validators, cli). `scripts/skills.py` stays as a thin façade that re-exports the full API and delegates the CLI, so every entry point is unchanged: the `skills.py {sync,generate,validate}` CLI, `import skills` from `bump_version.py`, and the tests that load it by path. Generation is byte-identical.

2. **Close the two guard gaps.** Derive each target's `plugin.json` `hooks` field from `hooks_render.out` and drop the redundant `targets.{codex,copilot,cursor}.hooks` keys, so the declaration and the generation output cannot disagree (one source, not two hand-maintained paths). Add `check_no_orphan_hook_scripts` so a `hooks/*.py` that is not wired into the source fails CI (previously only the reverse was checked). Also make `bump_version.py` regenerate routing + hook-wiring files, not just the plugin manifests + `manifest.json`. Adds 4 tests.

3. **Make the source obvious.** Move `plugin.meta.json` into a `metaplugin/` folder and add `metaplugin/README.md` that names the one source, lists every generated artifact, and calls out what stays hand-edited in place (hook scripts, commands, skills). Update the `META_FILE` path, the CI trigger glob (`metaplugin/**`), the release commit list, the generated-dir README markers, and the docs.

## Test plan

- [x] `python3 scripts/skills.py generate` produces no diff to any generated file (byte-identical at every step).
- [x] `python3 scripts/skills.py validate` passes (now also flags orphan hook scripts).
- [x] `python3 -m unittest discover -s tests -p '*_test.py'` — 96 tests pass.
- [x] `python3 scripts/bump_version.py <v>` regenerates plugins + routing + hooks + manifest, then reverted.
- [ ] CI (validate-manifest) green.


pkosiec (commented):
Reviewed the refactor — splitting `scripts/skills.py` into the `scripts/skillsgen/` package + a thin re-export façade, and moving `plugin.meta.json` into `metaplugin/`. The mechanics hold up well:

- `python3 scripts/skills.py validate` → "Everything is up to date" (no drift).
- The façade re-exports every `skills.<name>` the test suite uses (checked across all 6 test files), so the path-based test loader and `bump_version.py`'s `import skills` keep working.
- `validate-manifest.yml` runs the unit suite and triggers on `tests/**` / `scripts/**` / `metaplugin/**`, so the new orphan-hook / declared-hook tests execute in CI.

A few suggestions inline — only the first (release staging) is worth acting on before merge; the rest are minor/optional. Nothing blocks the refactor.

.github/workflows/release.yml
```diff
@@ -52,12 +52,12 @@ jobs:
...
-          # Stage all of them (unchanged files are a no-op) so a release can
-          # never commit a partial set, e.g. bump the Claude manifest but leave
-          # the Codex/Copilot ones behind.
+          # metaplugin/plugin.meta.json is the source; the rest are generated
+          # from it. Stage all of them (unchanged files are a no-op) so a release
+          # can never commit a partial set, e.g. bump the Claude manifest but
+          # leave the Codex/Copilot ones behind.
           git add \
-            plugin.meta.json \
+            metaplugin/plugin.meta.json \
```
pkosiec, line 62: **`bump_version.py` now regenerates more than this release step stages.**

`scripts/bump_version.py` also regenerates the routing files (`hooks/_routing_data.json`, `rules/databricks-routing.mdc`, `rules/README.md`) and the four hook-wiring files (`hooks/hooks.json`, `codex-hooks.json`, `copilot-hooks.json`, `cursor-hooks.json`), but this `git add` list still stages only the plugin/marketplace JSON + `manifest.json`.

Benign today (the version doesn't flow into those files, so the regen is a no-op), but it breaks the "never commit a partial set" invariant this comment asserts: if any of those generated files were ever drifted on `main`, a release would silently fix them on the runner, pass `validate`, publish the tag, and leave the fix uncommitted — and post-merge CI on `main` would then fail.

Fix (pick one): **(a)** add the routing + hook-wiring files to this `git add` list (keeps "regenerate full set → stage full set" symmetry), or **(b)** drop the routing/hook regen from `bump_version.py` since a version bump can't change their content.

scripts/skillsgen/plugins.py
```diff
@@ -0,0 +1,267 @@
...
+    in plugin.meta.json (so it gets no plugin keyword) and this fails, instead
+    of the keyword silently going missing from every target's plugin.json.
+    """
+    errors: list[str] = []
+    meta_skills = meta.get("skills", {})
+    disk_skills = {d.name for d in iter_skill_dirs(repo_root)}
+
+    for name in sorted(disk_skills - set(meta_skills)):
+        errors.append(
+            f"Stable skill '{name}' has no entry in plugin.meta.json \"skills\". "
```
pkosiec: **This error message points contributors at the old path.** It (and the sibling messages) hard-code the bare `plugin.meta.json`, but the file now lives at `metaplugin/plugin.meta.json`, so someone who hits this on `validate` is sent to the repo root. `cli.py` already interpolates the `META_FILE` constant — these per-check strings should too.

Other occurrences: `plugins.py:254,260`, `routing.py:144`, `hooks.py:171`, `validators.py:312,402`.

scripts/skillsgen/validators.py
```diff
@@ -0,0 +1,530 @@
...
+        errors.append(
+            '.codex-plugin/plugin.json must declare "hooks": "./hooks/codex-hooks.json". '
+            "Without it Codex defaults to hooks/hooks.json, the Claude Code wiring."
+        )
+    codex_hooks = repo_root / "hooks" / "codex-hooks.json"
+    if not codex_hooks.exists():
+        errors.append("Missing hooks/codex-hooks.json.")
+    else:
+        try:
+            cfg = json.loads(codex_hooks.read_text())
```
pkosiec: **DRY: this duplicates `_check_hook_wiring`.** `_check_hook_wiring` (L164) already parses a wiring file and verifies its `hooks/*.py` refs exist, returning the parsed cfg — but `check_codex_plugin` here (and `check_copilot_plugin`, ~L424) re-implement the parse + `re.findall(r"hooks/…\.py")` + existence loop inline. Both could route through `_check_hook_wiring` and layer the version / event-name checks on its return value. Pre-existing; carried over by the split — optional.

scripts/skillsgen/validators.py
```diff
@@ -0,0 +1,530 @@
...
+            '.codex-plugin/plugin.json "name" must be "databricks" (the install '
+            "identifier; the marketplace entry and install docs key on it)."
+        )
+
+    claude_path = repo_root / ".claude-plugin" / "plugin.json"
+    try:
+        claude_version = json.loads(claude_path.read_text()).get("version")
+    except Exception:
+        claude_version = None
+    if claude_version and plugin.get("version") != claude_version:
```
pkosiec: **Likely-redundant cross-version check (minor).** Now that all four `plugin.json` are generated from the single `meta["version"]` and `check_generated_plugins` enforces byte-exact equality, any version mismatch is already caught by the drift check. This Claude-vs-Codex/Copilot comparison is harmless defense-in-depth but could be dropped to shrink the surface.

pkosiec (approved):

.github/workflows/release.yml
simonfaltum: Went with (a) plus a guard. The `git add` list now also stages the routing files (`hooks/_routing_data.json`, `rules/databricks-routing.mdc`, `rules/README.md`) and the four hook-wiring files (`hooks/hooks.json`, `codex-hooks.json`, `copilot-hooks.json`, `cursor-hooks.json`).

On top of that, after staging I run `git diff --quiet` and fail the release if any tracked file is still modified. So if a new generated artifact is ever added without updating this list, the release fails loudly instead of publishing a tag with the regen left uncommitted (which is the failure mode you described). Keeps the regenerate-full-set / stage-full-set symmetry and makes it self-enforcing. (5252dec)

scripts/skillsgen/plugins.py
simonfaltum: Fixed in beff450. Interpolated `META_FILE` in every per-check message: skill-coverage (`plugins.py`), routing coverage + pattern (`routing.py`), orphan-hook (`hooks.py`), and the generated-file drift message in `common.py`. The two `validators.py` version-mismatch strings you flagged are gone entirely, since I dropped that cross-check per the comment below (3e3125d).

scripts/skillsgen/validators.py
simonfaltum: Done in 3e3125d. `check_codex_plugin` and `check_copilot_plugin` now route through `_check_hook_wiring` and layer the copilot version / event-name checks on its return value, matching how `check_cursor_plugin` already worked.

scripts/skillsgen/validators.py
simonfaltum: Dropped both (codex + copilot) in 3e3125d. `check_generated_plugins` enforces byte-exact equality against the meta-generated files, and the version flows from `meta["version"]` into all four targets, so a mismatch is already caught by the drift check, which is also strictly stronger than the old Claude-vs-target comparison.

PR #157, opened 2026-06-17
skills(experimental): Real-Time Mode bug fixes + two new references

DanJKeeling:
## Summary

Fixes incorrect RTM facts in two existing references and adds two new references covering RTM essentials and the Lakebase `foreach` sink — the canonical sink shape for a realtime app whose serving layer is Lakebase.

Scope: `experimental/databricks-spark-structured-streaming/` only. No changes to `skills/`.

### `references/trigger-and-cost-optimization.md`

The RTM coverage made several claims that don't match current docs and wouldn't run as written:

- `.trigger(realTime=True)` — boolean form doesn't exist in PySpark; the duration string is required
- "Photon enabled" — Photon must be **off** for RTM
- "Databricks 13.3+" — current minimum is DBR 16.4 LTS
- "Latency: < 800ms" — current docs cite "as low as 5 ms"

Five instances of these claims across the file are corrected, including the Trigger Selection Guide table, the "Real-Time Mode (RTM) Configuration" section, the cost-strategy block, and the "RTM not working" Common Issues row.

### `references/kafka-streaming.md`

Three patterns were broken as written:

- **Pattern 4 (Event Enrichment)**: used `.trigger(realTime=True)` (boolean) AND `.join(user_dim, ...)` without `broadcast()` — RTM only supports broadcast stream-static joins, so it wouldn't have run under RTM. Fixed to wrap with `broadcast()`, add `outputMode("update")`, use the correct `realTime="5 minutes"` form, and import `broadcast` in the pattern's code block.
- **Patterns 5 & 6 (Multi-Topic Routing, Schema Validation with DLQ)**: both used `foreachBatch(...)` combined with `.trigger(realTime=True)`, directly contradicting the same file's Pattern 3 correct note that "forEachBatch is NOT supported in RTM". Switched both to `.trigger(processingTime="30 seconds")` (consistent with the file's Quick Start) with an inline comment on the `foreachBatch` line.

Also fixed: Performance Tuning table row, Common Issues "High latency" row, and Production Checklist row that repeated the `< 800ms` framing.

### `references/real-time-mode.md` (175 lines)

The must-know RTM facts in one place:

- Cluster setup constraints (DBR 16.4 LTS+, Classic compute only, autoscaling/Photon/spot OFF, the `spark.databricks.streaming.realTimeMode.enabled` flag, Dedicated mode for UDFs)
- Slot math (total worker vCPUs ≥ sum of partitions across all stages; the `[CONCURRENT_SCHEDULER_INSUFFICIENT_SLOT]` error)
- Trigger and output mode (the `"5 minutes"` is checkpoint interval, not trigger interval)
- Supported sources/sinks matrix
- `transformWithState` per-row behavior (vs micro-batch per-key-per-batch) — the most commonly-missed RTM semantic difference
- Verifying and observing RTM (physical-plan check for `RealTimeStreamScan`; the three built-in latency metrics with the sink-write caveat; diagnose-by-metric decision tree)
- Delivery semantics (exactly-once within Spark, at-least-once for Kafka and `foreach`)
- Common error class table

### `references/lakebase-sink-python.md` (255 lines)

A duck-typed Python `foreach` sink that writes streaming records into Lakebase Postgres — the canonical sink shape for an RTM realtime app whose serving layer is Lakebase. Cross-links to the stable `databricks-lakebase` skill for general Lakebase mechanics (auth methods, CU-to-connections cap, scaling) rather than restating them:

- Why this pattern vs `foreachBatch + df.write.jdbc()`
- Two non-obvious rules: duck-typed class (no `ForeachWriter` Python base class — that symbol is Scala-only), and native Postgres password auth (not OAuth) on executors
- Full `LakebaseSink` class with time-based commits (`max_dwell_ms`), `INSERT … ON CONFLICT` idempotent upserts, per-partition lifecycle
- Wiring example, delivery semantics (cross-link to `real-time-mode.md`), `max_dwell_ms` tuning, connection budget formula tying app-pool size + executor connections to the CU cap
- Target table DDL
- **Performance & longevity** section: dead-tuple bloat on hot-key upserts (per-table autovacuum tuning), the CDF-on-upsert-table anti-pattern (replication slot pins xmin horizon → unbounded bloat), and the reconnect-instead-of-fail pattern for transient disconnects

## SKILL.md

Two new nav rows in the Core Patterns table pointing at the new references.

## Manifest

`manifest.json` regenerated by `python3 scripts/skills.py generate` to include the two new reference files. Only the `databricks-spark-structured-streaming` skill's file list changes.

## Validation

`python3 scripts/skills.py validate` → "Everything is up to date."

## Orthogonality

- No duplication with `stream-static-joins.md` (general broadcast pattern stays there; RTM file states the RTM-specific constraint only)
- No duplication with `stateful-operations.md` (general watermarks/state stores stay there; my `transformWithState` content is RTM-specific per-row behavior)
- No duplication with `streaming-best-practices.md` (general checklist stays there; my RTM file states stricter hard requirements)
- No duplication with `multi-sink-writes.md` (that's `foreachBatch`; my file is `foreach` — explicitly contrasts)
- Cross-links into `skills/databricks-lakebase/references/connectivity.md`, `computes-and-scaling.md`, and `lakehouse-sync.md` rather than restating their content


experimental/databricks-spark-structured-streaming/references/kafka-streaming.md
```diff
@@ -171,12 +171,14 @@ query = (enriched_df
 Enrich events with dimension data:
```
peterst28, line 171: Shouldn't we recommend RTM for kafka->kafka streams?  I think it's better than microbatch for this case.

experimental/databricks-spark-structured-streaming/references/kafka-streaming.md
```diff
@@ -231,8 +234,8 @@ def route_events(batch_df, batch_id):
         .save()
 
 parsed_df.writeStream \
-    .foreachBatch(route_events) \
-    .trigger(realTime=True) \
+    .foreachBatch(route_events) \                    # foreachBatch is NOT supported in RTM — use processingTime
```
peterst28, line 237: Comment is probably not necessary here, although it's useful for the reviewer. :)

experimental/databricks-spark-structured-streaming/references/kafka-streaming.md
```diff
@@ -281,8 +284,8 @@ def validate_and_route(batch_df, batch_id):
             .save()
 
 source_df.writeStream \
-    .foreachBatch(validate_and_route) \
-    .trigger(realTime=True) \
+    .foreachBatch(validate_and_route) \              # foreachBatch is NOT supported in RTM — use processingTime
```
peterst28, line 287: Same.  Comment not necessary.

experimental/databricks-spark-structured-streaming/references/kafka-streaming.md
```diff
@@ -352,7 +355,7 @@ df.writeStream \
 | minPartitions | Match Kafka partitions | Optimal parallelism |
 | maxOffsetsPerTrigger | 10,000-100,000 | Balance latency vs throughput |
 | trigger interval | Business SLA / 3 | Recovery time buffer |
-| RTM | Only if < 800ms required | Microbatch more cost-effective |
+| RTM | When sub-second E2E latency is required | Micro-batch is more cost-effective for second-or-longer SLAs |
```
peterst28: Recommend RTM for kafka -> kafka?

experimental/databricks-spark-structured-streaming/references/kafka-streaming.md
```diff
@@ -402,7 +405,7 @@ for stream in spark.streams.active:
 - [ ] Checkpoint location is persistent (UC volumes, not DBFS)
 - [ ] Unique checkpoint per pipeline
 - [ ] Fixed-size cluster (no autoscaling for streaming/RTM)
-- [ ] RTM enabled only if latency < 800ms required
+- [ ] RTM enabled only when sub-second E2E latency is required
```
peterst28: Too strong?  Good for kafka -> kafka

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
```diff
@@ -0,0 +1,175 @@
+---
+name: real-time-mode
+description: Real-Time Mode (RTM) for Spark Structured Streaming on Databricks — sub-second end-to-end latency. Use when building realtime apps, low-latency operational pipelines, or any streaming workload with SLAs measured in milliseconds rather than seconds.
```
peterst28: or kafka -> kafka!

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
```diff
@@ -0,0 +1,175 @@
...
+.trigger(realTime="5 minutes")
+.outputMode("update")
+```
+
+**Scala:**
+```scala
+import org.apache.spark.sql.execution.streaming.RealTimeTrigger
+import org.apache.spark.sql.streaming.OutputMode
+
+.trigger(RealTimeTrigger.apply())          // or RealTimeTrigger.apply("5 minutes")
```
peterst28: I think it's better to use the explicit "5 minutes"

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
```diff
@@ -0,0 +1,175 @@
...
+.outputMode(OutputMode.Update())
+```
+
+Two things people get wrong:
+
+1. **The `"5 minutes"` is the checkpoint interval, not the trigger interval.** RTM is continuous; it doesn't "fire" on a schedule. The duration controls how often the engine pauses to checkpoint. Set it to **at least 5 minutes** — shorter intervals cause frequent multi-second pauses; longer intervals mean more potential reprocessing on restart. Inter-batch time should stay ≤3 seconds (≤1% of the batch duration) or P99 latency rises.
```
peterst28: It's ok to set a 1 minute interval for debugging/testing to get stats quickly.

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
```diff
@@ -0,0 +1,175 @@
...
+|---|---|
+| Stateless: Kafka source (`maxPartitions=8`) → Kafka sink | 8 |
+| + one shuffle (group by, dedup) with `spark.sql.shuffle.partitions=20` | 8 + 20 = 28 |
+| + one explicit `.repartition(20)` | 8 + 20 + 20 = 48 |
+
+If `maxPartitions` is unset, the source partition count equals the Kafka topic's partition count. If under-sized, the query throws `[CONCURRENT_SCHEDULER_INSUFFICIENT_SLOT]` and stalls or fails.
+
+## Supported sources and sinks
+
+| Source / sink | As source | As sink |
```
peterst28, line 63: Do you need to explain what the symbols mean?  Or replace '-' with N/A?

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
```diff
@@ -0,0 +1,175 @@
...
+| Auto Loader / `cloudFiles` | ✗ | — |
+| Files / object storage directly | ✗ | — |
+| Google Pub/Sub | ✗ | ✗ |
+| Apache Pulsar | ✗ | ✗ |
+| Custom sink via `foreach` (Python class or Scala `ForeachWriter`) | — | ✓ |
+| `foreachBatch` | — | ✗ |
+
+**File-based sources (Auto Loader, direct file reads, Delta) are NOT supported in RTM.** They belong to micro-batch streaming. If your data lives in files and you need sub-second latency, ingest into Kafka / Event Hubs first.
+
+For the `foreach` sink to Lakebase Postgres, see [lakebase-sink-python.md](lakebase-sink-python.md).
```
peterst28: Or as an example for how to build a custom sink.

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Not supported in RTM
+
+- Watermark-based windowed aggregations
+- Stream-stream joins
+- `flatMapGroupsWithState` (older API)
+- `foreachBatch` and `mapPartitions`
+- Output modes `append` and `complete`
+
+## `transformWithState` behavior change
```
peterst28, line 105: May be worth pointing out that transformWithState is incredibly useful in RTM b/c it helps get around the many limitations.

peterst28: I don't see any mention of transformWithState in this repo.  Maybe worth going into that in your next pull request.  I might actually remove this section until you've added info about transformWithState...

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
```diff
@@ -0,0 +1,175 @@
...
+If you see `MicroBatchScan` instead, the query is not running in RTM — check that the source is supported and the cluster Spark conf is set.
+
+### Built-in latency metrics
+
+Every RTM batch emits three latency metrics in `StreamingQueryProgress` under the `latencies` field. Per-batch percentile distributions (P0, P50, P90, P95, P99):
+
+| Metric | What it measures |
+|---|---|
+| `processingLatencyMs` | Read-to-write inside the query — how long the pipeline takes to process a record |
+| `sourceQueuingLatencyMs` | Source-append-time (e.g. Kafka log append) to first read by the query. Captures inter-batch time, message-bus queuing, upstream batching. |
```
peterst28, line 143: Caution: This number and e2eLatencyMs will be long if you're processing records that have been sitting in Kafka.  This is an easy mistake to make, I think.

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
```diff
@@ -0,0 +1,175 @@
...
+| `e2eLatencyMs` | Source-append-time to processed downstream. End-to-end. |
+
+**Caveat: `e2eLatencyMs` does not currently include the sink write time.** If perceived latency is higher than `e2eLatencyMs` suggests, the gap is in the sink.
+
+Read these metrics via a `StreamingQueryListener` or by inspecting `query.lastProgress`.
```
peterst28: Users can see it in the UI.  Worth mentioning?

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Diagnose-by-metric
+
+| Symptom | Likely cause | Where to look |
+|---|---|---|
+| High `processingLatencyMs` | Slow operator inside the query | Per-stage metrics (set `spark.databricks.streaming.execution.enableDebugMetrics = true`); CPU profile |
+| High `sourceQueuingLatencyMs` | Inter-batch time too long, or upstream source latency | Inter-batch time in driver logs; Kafka `kafka.fetch.max.wait.ms` (default 500 ms — drop to 50 for low latency); upstream batching |
+| `e2eLatencyMs` looks fine but the app feels slow | Sink write time (not in `e2eLatencyMs`) | Measure sink flush duration directly |
+| Latency climbing over time | Memory pressure or GC growing | Executor stdout for Full GC events (long `user` CPU times); cluster restart as immediate mitigation |
```
peterst28, line 159: Oh boy.  This is a complicated topic.  We did a whole talk on it!

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Delivery semantics
+
+RTM preserves Structured Streaming's standard fault-tolerance guarantees:
+
+- **Exactly-once within Spark.** Operators, state stores, and supported sinks are all exactly-once.
+- **At-least-once for Kafka and `foreach` sinks.** This is the boundary case to watch — anywhere data leaves Spark via `foreach` (custom writers, side effects) or is written to Kafka, the same record may be delivered more than once on restart or task retry. Design custom sinks to be idempotent.
```
peterst28: Kafka is at least once?  Is that documented somewhere?

experimental/databricks-spark-structured-streaming/references/lakebase-sink-python.md
```diff
@@ -0,0 +1,257 @@
...
+);
+
+CREATE INDEX IF NOT EXISTS events_ts_idx ON events (timestamp DESC);
+```
+
+The primary key is what `ON CONFLICT` upserts against. The descending timestamp index supports the "give me the latest N" query a downstream app will run. Add business columns (metric, category, payload JSON) as your app needs.
+
+## Performance & longevity
+
+The sink performs well for moderate rates and short-lived or append-mostly workloads. The case that needs care is the one it's most often used for — **continuously upserting a small set of hot rows** to serve a live app — where dead-tuple churn can degrade it over a long run.
```
peterst28: Do you want to discuss the upsert approach that uses Lakebase as a state store, or do you consider that bad practice?

peterst28 (approved):
Fantastic improvement for realtime mode.  The existing skills for realtime are anemic and often just plain wrong.  My comments are mostly nitpicks and suggestions for future improvements.

apingledbx:
Things I would add: 
SDP on RTM PuPr 
Postgres sink in PuPr 
Stream-Stream inner join in PrPr (mention it's on the roadmap and teams can reach out to their databricks account team) 
Adaptive Stream Repartitioning (autoscaling for Stateless queries on RTM SDP)

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
apingledbx: Kafka sink is not idempotent so delivery guarantee is atleast once. Just like in MBM. Processing guarantees are exactly once - meaning each record will affect state only once. That's true for both RTM and MBM

QuentinAmbard:
## Review summary

**Verdict: fix-then-merge.** No P0 blockers. Four P1 findings to address; the rest is solid work that materially improves the skill. The core motivation — fixing the bogus `realTime=True`, `Photon enabled`, `<800ms`, and DBR 13.3 claims — is correct on every count. docs.databricks.com confirms: DBR 16.4 LTS minimum, Photon OFF, "as low as 5 ms" latency, `outputMode("update")` only, `realTime="5 minutes"` string syntax, exact Spark conf name. The `transformWithState` per-row claim is also confirmed by docs.

---

### P1 — should fix before merge

**1. Cross-tier links don't follow the install-flatten convention.** `lakebase-sink-python.md` lines 10, 22, 184, 234 use `../../../skills/databricks-lakebase/references/<file>.md`. Every other experimental→stable cross-skill link in the repo uses the install-flatten form (`../<sibling>/...`) — at install time the `experimental/` and `skills/` trees collapse into one directory. The 3-dot links resolve on disk in-repo but will break post-install. Fix: rewrite to `../databricks-lakebase/references/connectivity.md`, `computes-and-scaling.md`, `lakehouse-sync.md` (4 instances).

**2. RTM cluster requirements duplicated between `real-time-mode.md` and `trigger-and-cost-optimization.md`.** The new file is meant to be the deep treatment, and `trigger-and-cost-optimization.md` Pattern 3 correctly says "see real-time-mode.md for the deep treatment." But the `## Real-Time Mode (RTM) Configuration` section below (lines ~300-340) still restates DBR 16.4 LTS, Photon OFF, autoscaling/spot OFF, the Spark conf, slot math, output-mode requirement, trigger syntax. Two copies will drift the moment one gets updated. Fix: collapse that section to a 2-line pointer to `real-time-mode.md`; keep only the cost-specific framing ("Cost: continuous cluster, Photon disabled").

**3. The "5 minutes = checkpoint interval, not trigger interval" framing is overstated.** Databricks' own docs are inconsistent: the Scala comment on the setup page does call it the "checkpoint interval," but the Python comment calls it "the interval" (unqualified), and the RTM overview page calls it "long-running batches (the default is five minutes)" — i.e. batch duration. The PR picks the Scala phrasing as definitive across 5 occurrences (`real-time-mode.md`:56, `trigger-and-cost-optimization.md` lines 63/121/304/429, `lakebase-sink-python.md` wiring example). Fix: soften to "this is the long-running batch duration, sometimes called the checkpoint interval — set to ≥ 5 minutes" with a doc link, rather than asserting "checkpoint interval, NOT trigger interval" as the corrected fact. The ≥ 5 min floor stands either way; only the framing needs the hedge.

**4. Four bracketed RTM error class names cited without evidence.** The error-class table at the end of `real-time-mode.md` lists `[CONCURRENT_SCHEDULER_INSUFFICIENT_SLOT]`, `[STREAMING_REAL_TIME_MODE.OUTPUT_MODE_NOT_SUPPORTED]`, `[STREAMING_REAL_TIME_MODE.OPERATOR_OR_SINK_NOT_IN_ALLOWLIST]`, `[STREAMING_REAL_TIME_MODE.INPUT_STREAM_NOT_SUPPORTED]`. Could not verify these exact bracketed names against the overview, setup, or performance docs pages. Fix: either attach a captured error message from a real RTM job that produced one (the shapes are realistic Spark error classes), or soften the heading to "Common errors and their causes" without asserting the literal bracketed name. Agents will grep for these strings.

### P2 — nice to improve

**5. `InterfaceError` in the reconnect catch is over-broad.** `lakebase-sink-python.md` § "Reconnect instead of failing" catches `(psycopg.OperationalError, psycopg.InterfaceError)`. Per psycopg3 docs, `OperationalError` is the correct class for unexpected disconnects; `InterfaceError` is for interface-level bugs that won't be cured by a reconnect. Not wrong (it'll still recover), but tighter to drop `InterfaceError` unless the author has seen it from a real disconnect in practice.

**6. Three secondary claims worth a citation** (not findings per se — attach evidence and they close):
- Lakebase **create-time** "native Postgres password login enabled" flag (`lakebase-sink-python.md` prerequisites) — not found on the Lakebase get-started doc.
- "REPLICA IDENTITY FULL doubles WAL per update" magnitude — the Postgres behavior is real (full old + new row vs PK only), but "doubles" is an approximation; a postgresql.org link closes it.
- `spark.sql.execution.sortBeforeRepartition = false` ↔ RTM low-latency linkage — the conf exists in Spark source, but no Databricks page ties it to RTM specifically.

---

### Checked and clean

- `python3 scripts/skills.py validate` passes; manifest regenerated, not hand-edited.
- Every relative link in changed files resolves on disk. All 11 references (incl. 2 new) are linked from SKILL.md (no orphans).
- No stale `realTime=True`, `< 800ms`, `Photon enabled`, or "DBR 13.3" claim survives anywhere in the skill — thorough sweep across all touched files; untouched references contain no contradicting RTM facts.
- Cross-skill targets exist: `skills/databricks-lakebase/references/{connectivity,computes-and-scaling,lakehouse-sync}.md`.
- Verified against `docs.databricks.com`: DBR 16.4 LTS minimum, Photon OFF, `5 ms` latency floor, `outputMode("update")` requirement, `realTime="5 minutes"` string syntax (boolean form doesn't exist), `spark.databricks.streaming.realTimeMode.enabled` exact conf name, Classic-compute-only restriction, `transformWithState` per-row behavior in RTM, supported sources/sinks matrix.
- Verified locally: `pyspark.sql.streaming.ForeachWriter` is not a Python-exported symbol (PySpark 3.5.2 has no such export) — the duck-typed-class claim is correct.
- Verified against `skills/databricks-lakebase/references/computes-and-scaling.md`: CU_1 ≈ 209, CU_4 ≈ 839 connection-cap numbers match.
- Verified against `docs.databricks.com/aws/en/oltp/projects/get-started`: default DB name `databricks_postgres` is correct.
- Verified against `docs.databricks.com/aws/en/oltp/projects/quickstart-lakebase-cdf`: `REPLICA IDENTITY FULL` is the documented requirement.
- Verified against `postgresql.org`: replication-slot xmin pinning blocking autovacuum is standard Postgres behavior; `autovacuum_vacuum_cost_delay = 0` disables cost-based throttling.
- No invented absolutes — `MUST/NEVER/only` usages all reflect real RTM constraints (output-mode enforced by error class, file-sources legitimately unsupported, etc.).
- Whole-skill coherence pass surfaced no pre-existing findings beyond the duplication already noted in P1 #2.

experimental/databricks-spark-structured-streaming/references/kafka-streaming.md
DanJKeeling: **✅ Done in pass 2 (commit 8f43bbb) — see [follow-up reply](https://github.com/databricks/databricks-agent-skills/pull/157#discussion_r3473626363) below.**

experimental/databricks-spark-structured-streaming/references/kafka-streaming.md
DanJKeeling: Pushing back gently — keeping the inline comment. The skill's audience is coding agents; if an agent generating routing code reads only this pattern, the inline note prevents them from generating `foreachBatch + realTime` without re-reading Pattern 3. Cost is one line; benefit is preventing a known-broken combination. Happy to move it into the lead-in prose for the pattern if you'd prefer that location.

experimental/databricks-spark-structured-streaming/references/kafka-streaming.md
DanJKeeling: Same response as :237 — keeping for the agent-discoverability reason. Same offer to move it into the lead-in prose if you prefer.

experimental/databricks-spark-structured-streaming/references/kafka-streaming.md
DanJKeeling: **✅ Done in pass 2 (commit 8f43bbb) — see [follow-up reply](https://github.com/databricks/databricks-agent-skills/pull/157#discussion_r3473626537) below.**

experimental/databricks-spark-structured-streaming/references/kafka-streaming.md
DanJKeeling: **✅ Done in pass 2 (commit 8f43bbb) — see [follow-up reply](https://github.com/databricks/databricks-agent-skills/pull/157#discussion_r3473626708) below.**

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
DanJKeeling: **✅ Done in pass 2 (commit 8f43bbb) — see [follow-up reply](https://github.com/databricks/databricks-agent-skills/pull/157#discussion_r3473626884) below.**

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
DanJKeeling: Fixed in pass 1 — the Scala line is now `RealTimeTrigger.apply("5 minutes")` (commit 49ce3ca).

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
DanJKeeling: Fixed in pass 1 — added 'For development or debugging, dropping to 1 minute is fine if you want to see metrics emit more frequently.'

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
DanJKeeling: Fixed in pass 1 — all 5 `—` instances in the sources/sinks table replaced with `N/A`. The ✓/✗ remained since they're conventional.

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
DanJKeeling: **✅ Done in pass 2 (commit 8f43bbb) — see [follow-up reply](https://github.com/databricks/databricks-agent-skills/pull/157#discussion_r3473627517) below.**

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
DanJKeeling: **✅ Done in pass 2 (commit 8f43bbb) — see [follow-up reply](https://github.com/databricks/databricks-agent-skills/pull/157#discussion_r3473627066) below.**

DanJKeeling: Pushing back gently on removing the section. The current treatment documents a single RTM-specific semantic difference (per-row vs per-key `handleInputRows` invocation) that's high-signal even without broader `transformWithState` coverage. Without it, agents writing stateful processors will produce code that's correct for micro-batch but silently broken under RTM. I'd prefer to keep the section and tackle a deeper `transformWithState` reference as a follow-up PR — possibly as a parent reference file this section then links into.

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
DanJKeeling: **✅ Done in pass 2 (commit 8f43bbb) — see [follow-up reply](https://github.com/databricks/databricks-agent-skills/pull/157#discussion_r3473627208) below.**

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
DanJKeeling: **✅ Done in pass 2 (commit 8f43bbb) — see [follow-up reply](https://github.com/databricks/databricks-agent-skills/pull/157#discussion_r3473627357) below.**

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
DanJKeeling: Acknowledged — agree it's a complicated topic. The current treatment is intentionally minimum-viable, and in pass 1 the bracketed error-class names were softened to symptom-based descriptions (some of the original names came from an internal doc that isn't appropriate to cite from a public skill). If you have a public link to that talk I'd reference it from this section.

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
DanJKeeling: Fixed in pass 1 — added the [Spark Structured Streaming fault-tolerance guarantees](https://spark.apache.org/docs/latest/structured-streaming-programming-guide.html#fault-tolerance-semantics) link plus rationale: 'Kafka writes aren't idempotent without producer-side guards, and `foreach`'s lifecycle has no exactly-once commit protocol.' @apingledbx confirmed this below.

experimental/databricks-spark-structured-streaming/references/lakebase-sink-python.md
DanJKeeling: Out of scope for this PR but a great topic. Lakebase-as-state-store sits at the boundary between sink (this file's scope) and state primitive (closer to `transformWithState` territory). Marking as a follow-up; happy to draft if there's appetite.

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
DanJKeeling: Thanks for the confirmation — used your framing in pass 1: 'Kafka writes aren't idempotent without producer-side guards.' Combined with the Spark fault-tolerance guarantees link.

DanJKeeling:
## Pass 1 remediation pushed

Thanks all for the thorough review. Force-pushed amendment **49ce3ca** covering the mechanical/citation fixes from this round. Per-comment replies above explain what landed against each inline comment, including push-backs on two items. Pass 2 (broaden RTM use cases to include kafka→kafka; transformWithState framing note; backlog-latency caveat; UI mention; reframe Lakebase-sink as a generic foreach example) is queued next.

### Addressing @QuentinAmbard's findings

The review summary was a top-level comment so doesn't get inline replies — addressing here:

| Item | Status |
|---|---|
| **P1.1** Cross-tier link form | ✅ Fixed. 4 instances in `lakebase-sink-python.md` collapsed from `../../../skills/...` to `../databricks-lakebase/...` to match the install-flatten convention. |
| **P1.2** Duplicated cluster requirements in `trigger-and-cost-optimization.md` | ✅ Fixed. The `## Real-Time Mode (RTM) Configuration` section collapsed to a 2-line pointer at `real-time-mode.md`; only the cost-specific framing kept. |
| **P1.3** "Checkpoint interval, not trigger interval" overstated | ✅ Softened in all 5 places to "long-running batch duration (sometimes called the checkpoint interval) — set to ≥ 5 minutes." |
| **P1.4** Bracketed RTM error class names | ✅ Softened. Common errors table now uses symptom descriptions; the inline mention of `[CONCURRENT_SCHEDULER_INSUFFICIENT_SLOT]` was also removed. Heading softened to "Common errors" with a note that exact class names vary across DBR versions. (The original bracketed names came from an internal doc that isn't appropriate to cite from a public skill — good catch.) |
| **P2.5** `InterfaceError` over-broad | ✅ Fixed. Reconnect catch reduced to `psycopg.OperationalError` only. |
| **P2.6a** "Native Postgres password login enabled at create time" | ✅ Softened. Now reads "must permit native Postgres password login. If the instance is configured to require OAuth only..." with cross-link to `databricks-lakebase` connectivity. Dropped the unverified "at create time" assertion. |
| **P2.6b** REPLICA IDENTITY FULL "doubles WAL per update" | ✅ Softened to "substantially increases WAL volume per update — it logs the full old row plus the full new row instead of just primary-key + changed columns" with a postgresql.org REPLICA IDENTITY link. |
| **P2.6c** `sortBeforeRepartition=false` ↔ RTM linkage | ✅ Citation added. This is documented in the [RTM examples doc](https://docs.databricks.com/aws/en/structured-streaming/real-time/examples) under the Repartition example ("Due to a current implementation limitation, you must set the Spark configuration `spark.sql.execution.sortBeforeRepartition` to `false`..."). Link added inline. |

### @apingledbx's wishlist

Great future-PR list — capturing here:

- **SDP on RTM (PuPr)** — out of scope for the current "custom Structured Streaming surface only" framing. A follow-up `sdp-rtm.md` reference would land cleanly when the public preview matures.
- **Postgres sink (PuPr)** — once it ships natively, `lakebase-sink-python.md` slims significantly (the `foreach` workaround it documents becomes a "before native sink" footnote).
- **Stream-Stream inner join (PrPr)** — currently in my "Not supported" list per docs. Will update when the preview ships.
- **Adaptive Stream Repartitioning** (autoscaling for stateless RTM SDP queries) — would be a great addition once GA.

Will open separate PRs as these become public. Thanks for the forward-looking pointers.

### Pass 2 (next push)

- Broaden RTM use case framing to include kafka→kafka pipelines (Peter, comments at `:171`, `:358`, `:408`, `real-time-mode.md:3`)
- `transformWithState` "useful for working around RTM limitations" note (Peter, `:105`)
- Backlog-latency caveat for `sourceQueuingLatencyMs` and `e2eLatencyMs` (Peter, `:141`)
- Streaming dashboard UI mention alongside `StreamingQueryListener` / `lastProgress` (Peter, `:146`)
- Reframe Lakebase-sink pointer as also a foreach-sink worked example (Peter, `:80`)

Push-backs (per inline replies):
- Inline `foreachBatch is NOT supported in RTM` notes in Patterns 5 and 6 of `kafka-streaming.md` — keeping for agent-discoverability reasons.
- Removing the `transformWithState` section pending a deeper reference — keeping the per-row vs per-key semantic note since agents will trip on it.

Deferred to follow-up PRs:
- Lakebase-as-state-store discussion
- Deeper `transformWithState` reference

experimental/databricks-spark-structured-streaming/references/kafka-streaming.md
DanJKeeling: Pass 2 applied (commit 8f43bbb): Pattern 4 was already converted to RTM in pass 1 (broadcast + realTime). Performance Tuning table and Production Checklist rows now frame RTM as the default for kafka→kafka, not gated only on sub-second SLAs.

experimental/databricks-spark-structured-streaming/references/kafka-streaming.md
DanJKeeling: Pass 2 applied (commit 8f43bbb): Performance Tuning row now reads 'Default for kafka→kafka pipelines; required when sub-second E2E latency matters.'

experimental/databricks-spark-structured-streaming/references/kafka-streaming.md
DanJKeeling: Pass 2 applied (commit 8f43bbb): Production Checklist row softened to 'RTM enabled for kafka→kafka pipelines and any workload needing sub-second E2E latency.'

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
DanJKeeling: Pass 2 applied (commit 8f43bbb): `real-time-mode.md` description now includes kafka→kafka pipelines as a primary use case.

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
DanJKeeling: Pass 2 applied (commit 8f43bbb): added a lead-in paragraph framing `transformWithState` as the main escape hatch for working around RTM's other restrictions — event-time-window-like behavior, custom join logic, complex multi-row state transitions. The per-row vs per-key semantic note stays as the single concrete gotcha.

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
DanJKeeling: Pass 2 applied (commit 8f43bbb): added a backlog caveat after the existing `e2eLatencyMs` sink-write caveat — 'backlogs inflate `sourceQueuingLatencyMs` and `e2eLatencyMs`. Both clocks start at source-append time. If a query starts against records that have been sitting in Kafka for hours, those metrics will report hours...'

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
DanJKeeling: Pass 2 applied (commit 8f43bbb): mention added — 'Read these metrics via a `StreamingQueryListener`, by inspecting `query.lastProgress`, or in the streaming dashboard UI (the same metrics surface there).'

experimental/databricks-spark-structured-streaming/references/real-time-mode.md
DanJKeeling: Pass 2 applied (commit 8f43bbb): pointer at the sink reference now adds 'That file also serves as a worked example of the per-partition `foreach` lifecycle — useful as a template even if you're writing to a non-Lakebase target.'

DanJKeeling:
## Pass 2 pushed

Force-pushed **8f43bbb** addressing the content additions queued from @peterst28's review. Per-comment follow-ups on the original threads explain what landed against each:

- **RTM as default for kafka→kafka** (4 comment threads): Performance Tuning row and Production Checklist row now frame RTM as the default for kafka→kafka, not gated only on sub-second SLAs; `real-time-mode.md` description updated. Pattern 4 was already converted to RTM in pass 1.
- **`transformWithState` framing**: added lead-in paragraph framing it as the main escape hatch for RTM's other restrictions (event-time-window-like behavior, custom join logic, complex multi-row state transitions).
- **Backlog-latency caveat**: added to `sourceQueuingLatencyMs` / `e2eLatencyMs` — both clocks start at source-append time, so a backlog inflates both for the duration of the catchup.
- **Streaming UI**: mentioned alongside `StreamingQueryListener` and `query.lastProgress` as a metrics surface.
- **Lakebase-sink-python pointer**: now flags the file as a worked example of the per-partition `foreach` lifecycle generally, not just Lakebase specifically.

Remaining items deferred to follow-up PRs per the inline discussion:
- Deeper `transformWithState` reference
- Lakebase-as-state-store discussion
- Future RTM features (SDP-RTM PuPr, Postgres sink PuPr, stream-stream inner join PrPr, Adaptive Stream Repartitioning) as they become public

DanJKeeling:
## Pass 3 pushed (commit f2a9fe4)

Restructured `lakebase-sink-python.md` to lead with the native `format("postgresql")` sink — Public Preview, documented at [Connect to Lakebase](https://docs.databricks.com/aws/en/structured-streaming/lakebase) — picking up on @apingledbx's wishlist item. The manual `foreach` sink stays as a documented fallback for pre-DBR-18, Public-Preview-unfriendly environments, customization (composing writes, conditional routing), or non-Lakebase targets where the duck-typed Python class doubles as a worked `foreach` lifecycle example.

### `lakebase-sink-python.md` new structure

```
## Prerequisites                                  (general + per-option specifics)
## Target table                                   (applies to both)
## Option A: Native Lakebase sink (preferred when available)
  ### Unity Catalog form (simplest)
  ### Explicit endpoint form (non-UC tables)
  ### Options
  ### Connection behavior
  ### Triggers and output modes
## Option B: Manual `foreach` sink (fallback / customization)
  ### Two non-obvious rules (manual sink only)
  ### The `LakebaseSink` class
  ### Wiring into a stream
  ### Tuning `max_dwell_ms`
## Delivery semantics                             (applies to both)
## Connection budget                              (applies to both)
## Performance & longevity                        (applies to both)
  ### Dead-tuple bloat on hot-key upserts
  ### Do NOT put Change Data Feed on a high-churn upsert table
  ### Reconnect instead of failing the query     (manual sink only)
```

### Native sink content sourced from the docs

- `format("postgresql")`, requirements (DBR 18+, Classic Dedicated/Standard), Public Preview status
- UC `.toTable()` form with auto-table-creation + PK-inferred `upsertkey`
- Explicit-endpoint form including `project_id.branch_id` shorthand
- Options table: `endpoint`, `dbtable`, `database`, `upsertkey`, `batchsize`, `batchinterval`, `checkpointLocation`
- `JDBC_STREAMING_SINK_INVALID_OPTIONS` error class for unknown options
- Connection behavior: 1 task per connection default + optional `spark.databricks.sql.streaming.jdbc.tasksPerConnection` tuning
- Automatic retries on transient JDBC errors (connection failures, deadlocks, rate limiting)
- Upstream backpressure when DB can't keep up
- Trigger and output mode support (and the `append` ≡ `update` behavior when a PK exists)

### Cross-file updates

- `real-time-mode.md` line 80: pointer to the sink reference now mentions both options
- `SKILL.md` Core Patterns row renamed from "Lakebase `foreach` Sink" to "Lakebase Sink" with the native+fallback framing
- Manifest unchanged (no new files added; only an existing reference reorganized)

### Compliance with @QuentinAmbard's standards

- Cross-tier links use `../databricks-lakebase/...` form (P1.1)
- No duplicated cluster requirements — RTM cluster setup deferred to `real-time-mode.md` (P1.2)
- No "checkpoint interval, not trigger interval" assertions (P1.3)
- No bracketed RTM error class names (P1.4) — the one bracketed name in the file (`JDBC_STREAMING_SINK_INVALID_OPTIONS`) is the documented sink error class
- `InterfaceError` not in reconnect catch (P2.5)
- Native password not "at create time" (P2.6a)
- REPLICA IDENTITY FULL not "doubles" (P2.6b)
- `sortBeforeRepartition` citation intact (P2.6c)
- No invented `MUST`/`NEVER` absolutes

`python3 scripts/skills.py validate` passes.

DanJKeeling:
@apingledbx — picked up the Postgres sink (PuPr) item in pass 3. `lakebase-sink-python.md` now leads with `format("postgresql")` as Option A; the manual `foreach` class becomes Option B (fallback). Details in the pass-3 summary above. The other items (SDP on RTM, Stream-Stream inner join, Adaptive Stream Repartitioning) remain future-PR territory as they reach public availability.

DanJKeeling:
Closing this PR — the work has moved to the internal Databricks source-of-truth repo where I can iterate alongside other RTM updates from the team. Thanks @peterst28, @QuentinAmbard, @apingledbx, @dustinvannoy-db for the thorough review here — all the substantive findings have been carried forward (the bug fixes, the cross-tier links, the softened framings, and the citations Quentin called out). The error-class section was also restored with the bracketed `STREAMING_REAL_TIME_MODE.*` names now that the public docs page for them is live.

Additionally folding in product changes since this PR was last updated (RTM GA, SDP-on-RTM Public Preview, RTM stream-stream inner join on DBR 18+).

Once the internal change lands, it'll flow back here through the nightly sync and ship in a release.

PR #171, opened 2026-06-22
fix(copilot): load hooks from provider root

### LINKED ISSUE #158: Copilot plugin hooks not working
When I try to add the copilot plugin in VS code it loads the Claude hooks in hooks/hooks.json, not the copilot-hooks ones, resulting in failed hooks warnings. 

Based on the docs I think the hooks.json needs to be in the plugin root, might that be the issue? https://code.visualstudio.com/docs/agent-customization/agent-plugins#_hook-file-location

50sotero:
## Summary
Moves the Copilot plugin hook wiring to the provider root so Copilot-format plugins auto-discover `plugins/databricks/copilot/hooks.json` instead of the stale nested `hooks/hooks.json` path. Also pins ref-capable marketplace catalogs to release tags now that Copilot auto-loads executable hooks.

## Changes
- Emit Copilot bundle hook wiring at root `hooks.json`
- Update generator validation and docs for the Copilot hook path
- Pin Claude/Codex/Copilot marketplace refs to `v{version}`
- Add regression coverage for root Copilot hook wiring and release-tag refs

## Test Plan
- `python3 scripts/skills.py validate`
- `python3 -m unittest tests.skills_generator_test tests.hooks_generator_test`
- LSP diagnostics clean on modified Python files

Fixes #158


simonfaltum:
Thanks for this, @50sotero, and thanks @DustinGustavT for the report. The hooks-root fix is exactly right.

I've split the Copilot hooks-root fix into #175 (you're credited as co-author) so it can land on its own. It keeps the generator and validator changes, the rename, and the regression test.

I left out the marketplace tag-pinning part (`ref_template: main` -> `v{version}` and the `ref: v0.2.6` catalog pins), for two reasons:

1. We've decided not to pin the catalogs to release tags. They stay tracking `main`, where the committed bundle lives.
2. As written it would also have undone this very fix: `v0.2.6` already exists and predates the change, so it still carries `copilot/hooks/hooks.json`. Pinning the catalogs to it would have kept serving the old (buggy) layout to fresh installs.

Closing this in favor of #175. Really appreciate the contribution, the core fix is landing because of it.

PR #144, opened 2026-06-12
fix(skills): stop dashboard-intent leaking away from databricks-aibi-dashboards

pkosiec:
## Problem

A routing A/B test (176 headless Claude Code sessions, 6 install conditions, 11 prompts × 3 trials) measured which skill is invoked when `databricks-aibi-dashboards`, `databricks-app-design` ([#125](https://github.com/databricks/databricks-agent-skills/pull/125)), and `databricks-apps` are all installed.

| Install condition | dashboard-intent → aibi |
|---|---|
| aibi alone (baseline) | 12/12 |
| aibi + app-design | 17/18 |
| **realistic bundle** (aibi + app-design + databricks-apps) | **14/18 (78%)** 🚨 |

The fear that `databricks-app-design` "steals" dashboard requests is mostly unfounded head-to-head — its "NOT for Lakeview" disclaimer holds, and explicit "AI/BI"/"Lakeview" phrasing routed correctly 100% of the time. **The real leak is `databricks-apps`**: its description lists "dashboards" as a trigger and its body hard-routes *any* data-displaying app to `databricks-app-design` → AppKit, never mentioning `databricks-aibi-dashboards`. So "sales dashboard with a region filter" went apps → app-design → AppKit in 3/3 trials, and aibi was never even in the consideration set (0 mentions across all funnel transcripts). Ambiguous prompts hit that funnel 8/9 times.

## Fix

- **`databricks-app-design`** — lead the description with "custom-code Databricks Apps (AppKit/React) screens" instead of bare "dashboards, KPI pages…", and add an explicit rule: a plain "create a dashboard" means a managed AI/BI (Lakeview) dashboard → `databricks-aibi-dashboards`. Mirrored in the "When to use / when NOT" section.
- **`databricks-apps`** — drop "dashboards" from the trigger list; add the same deference rule to the description and an "Is this even a Databricks App?" paragraph at the top of the body.
- **Codex metadata** — reword `databricks-app-design/agents/openai.yaml` `default_prompt` so it no longer leads with dashboards (same leak on the OpenAI side).
- Regenerate `manifest.json`.

## Result (validated empirically, not just proposed)

With both edits: dashboard-intent routing → **18/18 (100%)** aibi in the bundle, ambiguous prompts now go to aibi or a clarifying question, and `databricks-app-design` keeps **6/6** of its legitimate app-UX traffic — zero over-correction.

This pull request and its description were written by Isaac.

PR #177, opened 2026-06-22
fix(bundle): regenerate stale app-design/apps bundle (drift from #144)

simonfaltum:
## What

Re-run `skills.py generate` to resync the `plugins/databricks/` bundle copies for `databricks-app-design` and `databricks-apps`: 12 files (8 `SKILL.md` + 4 `openai.yaml`, across the claude/codex/copilot/cursor targets). No source or version changes.

## Why

#144 edited the source `SKILL.md` for those two skills but did not re-run `skills.py generate`, so the committed bundle copies (which are generated from the source) went stale. `skills.py validate` is currently failing on `main` as a result. Because `validate-manifest` is non-blocking, the red commit landed.

## How verified

`python3 scripts/skills.py validate` reports `Everything is up to date.` on this branch (it fails on `main` before this change). The diff is exactly the regenerated bundle copies, confirmed against a clean `generate` run.

### DIFF
--- a/plugins/databricks/claude/skills/databricks-app-design/SKILL.md
+++ b/plugins/databricks/claude/skills/databricks-app-design/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: databricks-app-design
 parent: databricks-apps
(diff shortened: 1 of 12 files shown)


simonfaltum:
Closing in favor of letting the next `release.yml` run reconcile this. It regenerates the bundle from current `main` source and stamps it with the version, so the app-design/apps drift clears at release time without a separate version-less commit. Root cause for the record: #144 edited the app-design/apps source SKILL.md without re-running `skills.py generate`.

PR #192, opened 2026-06-26
refactor(metric-views): fold metric-view-advisor into the skill + make it less opinionated

dipankarkush-db:
## Summary

Folds the experimental `databricks-metric-view-advisor` into the stable `databricks-metric-views` skill as the `metric-view-advisor/` sub-folder, and reworks it per maintainer feedback (Slack thread w/ @dustinvannoy-db, Quentin, @simonfaltum).

This supersedes the original tiny doc-fix this PR started as — folding the advisor into the parent **dissolves the install-dependency problem entirely** (it now ships with the parent skill; no separate install).

## Changes

- **Fold:** `experimental/databricks-metric-view-advisor/` → `skills/databricks-metric-views/metric-view-advisor/`. It's no longer a separate skill/manifest entry — its files fold into `databricks-metric-views` and ship with it.
- **Less opinionated (Quentin's feedback):** refactored from a rigid step-by-step interview (mandatory STOP after every micro-step) into an **information-driven** flow — state the info needed and *why*, let the agent decide whether to proceed / ask / fetch, and checkpoint only at consequential actions (schema create, deploy, replace/drop) or when the user asked to review. All capabilities preserved (input sources, identifiers, target-schema validation/creation, overlap discovery, suggestions, definitions, materialization, deploy, sample queries, references, examples).
- **`format` blocks allowed:** docs now document `format` (number/currency/percentage/date) as a supported YAML 1.1 field. Replaced the old "never use format — causes parse errors" guidance with "set a valid `type`; omit if unsure."
- **Housekeeping:** dropped the advisor's standalone Codex metadata + icons (the parent provides them); added a pointer from the parent `SKILL.md`; updated `experimental/README.md`; regenerated `manifest.json` + the `plugins/databricks/` bundle.

## Docs gap-review (advisor-scoped)

Reviewed current Databricks semantic-layer / metric-view docs against the advisor. Most newly-documented features (`one_to_many` joins, `rely`/`at_most_one_match`, window `offset`/`trailing`/`leading`, the `agg` alias, `fields` preferred over `dimensions`) belong to the **baseline YAML spec, which lives in the parent skill — out of scope for the advisor.** The one advisor-scoped discrepancy was the `format` guidance, fixed here.

## Flags for reviewers

- **Nested `SKILL.md`:** no other skill nests a `SKILL.md`. The generator only discovers top-level skills, so this sub-folder `SKILL.md` is content the parent points to (matches the "explicit invocation, not keyword" intent) — but there's no precedent, so please confirm the runtime loaders handle it as expected.
- **Parent-spec discrepancies (not changed here, flagging):** the advisor restates "window measures require `version: 0.1`" and "joins must be many-to-one" — current docs say window measures need YAML 1.1+/DBR 18.1+ and `one_to_many` joins now exist. These belong to the parent skill's spec, so I left them for a parent-scoped update.

Signed-off-by: Dipankar Kushari <dipankar.kushari@databricks.com>

## Update: advisor now defers baseline spec to the parent

After a docs cross-check, the advisor was restating baseline spec the parent owns — some now stale (e.g. "window measures require `version: 0.1`", "joins must be many-to-one"). Restating spec in two places is a drift risk, so the advisor now **defers to the parent for the spec** and keeps only its own heuristics/gotchas (Important Notes, Limitations, `yaml-reference.md`, `step-4`, `patterns.md`). No advisor capability lost — deferred spec lives in the parent's sections; advisor-only facts the parent doesn't cover (no Delta Sharing, no data profiling, ALTER VIEW comment behavior) were kept inline.

**Parent-side follow-up:** the parent's *Join Rules* don't yet document join **cardinality** (many-to-one / `one_to_many`) or **LEFT OUTER** semantics — those should be added there (parent's domain).


experimental/databricks-metric-view-advisor/SKILL.md
```diff
@@ -23,7 +23,7 @@ Create Unity Catalog metric views from your existing Databricks assets — gold/
...
 
 **Prerequisites:**
-1. **The parent `databricks-metric-views` skill** must be available and loaded (see the REQUIRED note above) — this is mandatory, not optional.
+1. **The parent `databricks-metric-views` skill** must be available and loaded (see the REQUIRED note above) — this is mandatory, not optional. It now lives in the **stable** `skills/` tree (it was promoted out of `experimental/`), so it is **a separate install from this advisor**: it installs by default with `databricks aitools install` (no `--experimental` flag) or explicitly via `databricks aitools install databricks-metric-views`. Installing this advisor on its own (`databricks aitools install databricks-metric-view-advisor --experimental`) does **not** pull the parent in — install it separately, or the advisor cannot produce correct definitions.
```
dustinvannoy-db:
```suggestion
1. **The parent `databricks-metric-views` skill** must be available and loaded — this is mandatory, not optional. It installs by default with `databricks aitools install` (no `--experimental` flag) or explicitly via `databricks aitools install databricks-metric-views`. Installing this advisor on its own (`databricks aitools install databricks-metric-view-advisor --experimental`) does **not** pull the parent in — install it separately, or the advisor cannot produce correct definitions.
```

experimental/README.md
```diff
@@ -39,6 +39,13 @@ databricks aitools install databricks-genie --experimental
 
 See the root [README](../README.md) for details on the stable install path.
 
+> **`databricks-metric-view-advisor` has a stable dependency — install it separately.**
+> The advisor requires the `databricks-metric-views` skill, which now lives in the
+> stable [`../skills/`](../skills/) tree (it was promoted out of `experimental/`).
+> Install it via the default `databricks aitools install` (no `--experimental` flag)
+> or explicitly with `databricks aitools install databricks-metric-views`. Installing
+> the advisor on its own does **not** pull the parent in.
+
```
dustinvannoy-db: I want to rephrase this one also, but really this change will need to go internally first then we will sync it here.

dustinvannoy-db (changes requested):

dipankarkush-db:
Superseded by the internal universe PR (databricks-eng/universe#2147478), which is the source of truth that syncs to this mirror. Closing this public-fork PR. Thanks @dustinvannoy-db / Quentin for the review.

